PR #22, opened 2015-01-09
Use log from crates.io

vhbit:
In-tree `log` was deprecated a long ago and now almost all crates from
crates.io are linked against external log. Building capnproto-rust in 
this conditions is always a pain as rustc complains about 
linking to 2 different logs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,3 +15,6 @@ keywords = ["protocol", "serialization"]
 
 name = "capnp"
 path = "src/capnp/lib.rs"
+
+[dependencies]
+log = "0.1"


dwrensha:
Thanks for pointing this out! I didn't realize it was a problem. I just pushed a fix that eliminates the dependency altogether: https://github.com/dwrensha/capnproto-rust/commit/a9a1ea10c758f7498b8f86e2ee002fddb3517039

vhbit:
@dwrensha hmm, standard macros are not the best thing available, but they allow overwriting logger, i.e. it is not always `stderr` - it might be a file logger, `syslog` logger and so on. It means that library client might see no errors at all as they are written to unexpected location.

dwrensha:
I was not aware that `error!()` could write anywhere other than stderr. Is that documented anywhere? I'm not finding anything here: http://doc.rust-lang.org/log/log/index.html

vhbit:
To be honest, I didn't know it too, someone pointed it in reddit where
logging was discussed, [here is it](
http://doc.rust-lang.org/log/log/fn.set_logger.html). All std macros use
that task logger.

dwrensha:
Interesting, thanks. Reopening...

dwrensha:
This pull request is no longer relevant, as we've switched to return-value-based error handling.
See [this post](http://dwrensha.github.io/capnproto-rust/2015/03/21/error-handling-revisited.html).

PR #26, opened 2015-02-21
Use PhantomData and PhantomFn everywhere necessary to compile with latest Rust

kemurphy:
ContravariantLifetime was deprecated and there were a bunch of unused type parameters that needed markers.  Tried to match style as best as possible; let me know if it needs tweaking.


dwrensha:
Thanks!

PR #30, opened 2015-03-18
impl ::std::error::Error for capnp::Error

danburkert:
This makes it easier to interoperate with the `capnp::Error` type.


dwrensha:
Cool, thanks!

PR #39, opened 2015-04-19
Implement compute_serialized_size_in_words

Vaelden:
This new function allows to compute the size a message will have when serialized.

It can be useful when doing asynchronous IO, where it might be needed to make sure that the entire message has arrived before starting to process it.

(Inspired from the C++ version computeSerializedSizeInWords)


dwrensha:
Thanks!

PR #40, opened 2015-04-19
Rework traits HasTypeId and HasStructSize

Vaelden:
UFCS has made possible to unambiguously call a "static" method on a trait without `Self` appearing in the parameters.

This patch replaces the old hack of putting a parameter `Option<Self>` and use the new notation instead.

A pull request for changing generated code will follow.


dwrensha:
Thanks!

PR #41, opened 2015-04-21
std::io::{Read, Write} based message serialization

danburkert:
This is an experiment to see if removing the custom IO types (`InputStream` and `OutputStream`) and relying only on the std lib IO types (`Read`, `BufRead`, and `Write`) is worthwhile.  This PR adds unit tests as well as quickcheck tests to make sure serialization is consistent.  The resulting packed serialization implementation is largely the same, however the non-packed serialization code has significantly less usage of unsafely.  In my opinion the API exposed here is simpler and more approachable from the perspective of someone familiar with Rust, but not as familiar with Capnproto.  I believe (although I haven't written any benchmarks), that the performance should be fundamentally the same.


dwrensha:
Wow! This is interesting and promising. I like fact that you were able to add a bunch of tests and still end up with a net reduction in lines of code. :)

I went ahead and got the benchmark working with these changes: https://github.com/dwrensha/capnpc-rust/tree/remove-capnp-io

In my measurements, the packed serialization cases slow down by 10-20%. I suspect that a lot of that is due to the lack of a `BufWrite` trait. With some work, I suspect we could get it down to more like 5%, and for simplifications like this, I think that might be a price worth paying.

One thing I feel somewhat uneasy about is our use of `::std::io::ErrorKind::other` for the "packed input did not end cleanly on a segment boundary" case. It seems like we might want that to be a custom `capnp::Error` rather than a `::std::io::Error`, and if we're using `std::io::Read`, we don't have the flexibility to use our own error type. With the current `capnp::io` stuff, we could change `capnp::io::InputStream::read()` to return a `capnp::Result<usize>` rather than a `::std::io::Result<usize>`. I've been on the fence about this for a while.

danburkert:
Rebased against master, removed use of lifetime elision, and squashed.  I'm looking into the performance regression.

danburkert:
Checking in with regards to the performance regression in packed serialization.

My first thought was that the checked array writes in [`PackedWrite::write`](https://github.com/danburkert/capnproto-rust/blob/0a9ba1ae0184acafae9571a371e9c04e0d9ca5bb/src/serialize_packed.rs#L226) were causing the slowdown, but removing them in afcd871 yielded only a very minor performance increase.

As you suspected, I think the issue is the increased number of calls to the inner writer in `PackedWrite::write`, which are necessary because a buffer is not assumed.  in [a](https://github.com/danburkert/capnproto-rust/blob/io/src/serialize_packed.rs#L269) [few](https://github.com/danburkert/capnproto-rust/blob/io/src/serialize_packed.rs#L285) [places](https://github.com/danburkert/capnproto-rust/blob/io/src/serialize_packed.rs#L318) individual bytes are getting written to the underlying `Write` object, which is translating into a `memmove` system call, which is significantly slower for a single byte write.  It's perhaps unfortunate that the Rust `Write` trait doesn't have a special case for single byte writes.

My next move is to explore introducing an internal buffer to `PackedWrite::write` and see whether that can restore the performance.

danburkert:
f2c1018 adds a stack allocated internal buffer to `PackedWrite::write`, and performance is close to the current implementation.  @dwrensha mind taking another look, and seeing if this is sufficient?

src/serialize.rs
```diff
@@ -180,3 +210,185 @@ pub fn compute_serialized_size_in_words<U : MessageBuilder>(message: &mut U) ->
...
+                     1,0,0,0,   // 1 length
+                     199,0,0,0, // 199 length
+                     0,0,0,0,   // 0 length
+                     1,0,0,0],  // 1 length
+                   &buf[..]);
+        buf.clear();
+    }
+
+    #[test]
+    fn check_serde() {
```
dwrensha: When I first read this code I was confused for a bit about what it had to do with [serde](https://github.com/serde-rs/serde). I think "round_trip" might be a more clear name.

src/serialize_packed.rs
```diff
@@ -210,123 +199,103 @@ impl <'a, R : BufferedInputStream> InputStream for PackedInputStream<'a, R> {
...
-impl <'a, W : BufferedOutputStream> OutputStream for PackedOutputStream<'a, W> {
-    fn write(&mut self, in_buf : &[u8]) -> ::std::io::Result<()> {
+impl <W> Write for PackedWrite<W> where W: Write {
+    fn write(&mut self, in_buf: &[u8]) -> io::Result<usize> {
         unsafe {
-            let (mut out, mut buffer_end) = self.inner.get_write_buffer();
-            let mut buffer_begin = out;
-            let mut slow_buffer : [u8; 20] = [0; 20];
+            let mut buf_idx = 0;
+            let mut buf = [0; 64];
```
dwrensha: Readability: this wants some kind of type annotation.

src/serialize_packed.rs
```diff
@@ -468,8 +419,26 @@ mod tests {
             &[8,0,100,6,0,1,1,2, 0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0, 0,0,1,0,2,0,3,1],
             &[0xed,8,100,6,1,1,2, 0,2, 0xd4,1,2,3,1]);
 
-        expect_packs_to(&ZEROES[0 .. 16], &[0,1]);
+        expect_packs_to(&[0; 16], &[0,1]);
         expect_packs_to(&[0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0], &[0,2]);
+    }
 
+    #[test]
+    fn check_serde() {
```
dwrensha: Again, I think "serde" is confusing here.

src/serialize_packed.rs
```diff
@@ -357,62 +326,48 @@ impl <'a, W : BufferedOutputStream> OutputStream for PackedOutputStream<'a, W> {
...
+    use quickcheck::{quickcheck, TestResult};
+
+    use {Word, MessageReader};
+    use message::ReaderOptions;
+    use serialize::test::write_message_segments;
+    use serialize_packed::{PackedRead, PackedWrite};
+    use super::read_message;
+    use util::read_exact;
+
+
```
dwrensha: What is this extra white space doing?

src/util.rs
```diff
@@ -0,0 +1,68 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
```
dwrensha: It'd be nice if we could avoid having a generic utility module like this. This code should live as close to possible to where it is used. I feel like `read_exact()` should go in `serialize.rs`, but I guess that runs into visibility problems -- then it couldn't be private. Maybe what you've done here is the best compromise.

I wonder whether libstd is ever going to get an `read_exact()` method?

src/serialize.rs
```diff
@@ -47,125 +49,153 @@ impl MessageReader for OwnedSpaceMessageReader {
...
+/// For optimal performance, `read` should be a buffered reader type.
+pub fn read_message<R>(read: &mut R, options: ReaderOptions) -> Result<OwnedSpaceMessageReader>
+where R: Read {
+    let (total_words, segment_slices) = try!(read_frame(read, options));
+    read_segments(read, total_words, segment_slices, options)
+}
 
-    let segment_count : u32 = first_word[0].get() + 1;
+/// Reads a stream frame from `read` and returns the total number of words across all
+/// segments, as well as the segment offsets.
```
dwrensha: I think using the word "frame" like this is confusing. To me, a "frame" includes the payload as well as the header. The standard Cap'n Proto jargon for what this function reads is the "segment table", though I suppose that "frame header" would be an acceptable name too.

src/serialize.rs
```diff
@@ -47,125 +49,153 @@ impl MessageReader for OwnedSpaceMessageReader {
...
     let segments = message.get_segments_for_output();
-    let table_size : usize = (segments.len() + 2) & (!1);
+    try!(write_frame(write, segments));
+    write_segments(write, segments)
+}
 
-    let mut table : Vec<WireValue<u32>> = Vec::with_capacity(table_size);
-    unsafe { table.set_len(table_size) }
+/// Writes a stream frame to `write`.
+///
```
dwrensha: As on the read side, s/frame/segment table/.

dwrensha:
This looks quite good! I just had a few minor nitpicks.

I made a few performance measurements, and it looks like I no longer have anything to complain about there. :)

src/serialize_packed.rs
danburkert: I have a habit of leaving whitespace between std imports, imports from within the crate, and imports from other crates.  I can take that out.

danburkert: Actually this style somewhat matches the one in the [style guide](https://github.com/rust-lang/rust-guidelines/blob/master/style/imports.md).

src/util.rs
danburkert: `read_exact` is used in both `serialize` and `serialize_packed`, so this is the only way to do it without either leaking the declaration in the public API or defining it twice.  `ptr_sub` was only used in `serialize_packed`, so I moved that out.

src/serialize_packed.rs
danburkert: I haven't actually changed this- let me know if you want me to take the spaces out.  I introduced this style to quite a few places.

dwrensha:
Nice. I'll merge this just as soon as I update capnpc-rust and capnp-rpc-rust to be compatible.

danburkert:
Thanks!  Let me know if you want any help with that.

Hoverbear:
Looks great! Just migrated Raft!

PR #43, opened 2015-05-26
write_message should take a const message

danburkert:
This makes it possible to write a message from multiple contexts (threads, callbacks) without a lock.

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -141,10 +141,31 @@ impl <'a> SegmentArrayMessageReader<'a> {
     }
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
 pub enum AllocationStrategy {
     FixedSize,
     GrowHeuristically
+}
+
+#[cfg(test)]
+impl ::quickcheck::Arbitrary for AllocationStrategy {
+    fn arbitrary<G: ::quickcheck::Gen>(g: &mut G) -> AllocationStrategy {
+        if ::quickcheck::Arbitrary::arbitrary(g) {
+            AllocationStrategy::FixedSize
+        } else {
+            AllocationStrategy::GrowHeuristically
+        }
+    }
+    fn shrink(&self) -> Box<Iterator<Item=AllocationStrategy>+'static> {
+        match *self {
+            AllocationStrategy::FixedSize => {
+                ::quickcheck::empty_shrinker()
+            },
+            AllocationStrategy::GrowHeuristically => {
+                ::quickcheck::single_shrinker(AllocationStrategy::FixedSize)
+            },
+        }
+    }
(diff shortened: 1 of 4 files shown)


dwrensha:
Thanks! I agree that it would be good if `write_message` didn't need a mut ref. Something seems to be broken about this pull request, though. When I run the capnpc-rust tests I get:

```
  Running target/debug/capnpc_test-c761ad1da9661982

running 21 tests
test tests::double_far_pointers ... ok
test tests::inline_composite_list_int_overflow ... ok
test tests::empty_struct_list_amplification ... FAILED
test tests::all_types ... ok
test tests::all_types_multi_segment ... ok
test tests::test_big_struct ... ok
test tests::test_blob ... ok
test tests::test_any_pointer ... ok
test tests::setters ... ok
test tests::test_complex_list ... ok
test tests::test_defaults ... ok
test tests::test_constants ... ok
test tests::test_prim_list ... ok
test tests::test_union ... ok
test tests::text_builder_int_underflow ... ok
test tests::test_struct_list ... ok
test tests::upgrade_struct ... ok
test tests::test_set_root ... ok
test tests::threads ... ok
test tests::test_writable_struct_pointer ... ok
test tests::void_list_amplification ... FAILED

failures:

---- tests::empty_struct_list_amplification stdout ----
	thread 'tests::empty_struct_list_amplification' panicked at 'assertion failed: `(left == right)` (left: `4552964144`, right: `3`)', test.rs:736


---- tests::void_list_amplification stdout ----
	thread 'tests::void_list_amplification' panicked at 'assertion failed: `(left == right)` (left: `4676687920`, right: `2`)', test.rs:711
```

danburkert:
Yep, this is subtly broken.  I am adding some unit tests right now, and I'll push a fix soon.

danburkert:
OK the fix is in.  I ended up having to change the the interface of `get_segments_for_output` to return a `Vec<&[Word]>` instead of a `&[&[Word]]`.  This means that in the single-segment fast-path, `get_segments_for_output` will require allocating a 1-element vec whereas before it didn't.  I think this is worthwhile for the increase in functionality.  This change also makes the `BuilderArena` type slightly smaller (removes two fields) and simplifies the `get_segments_for_output` logic.

dwrensha:
I had a go at this myself and came up with this: https://github.com/dwrensha/capnproto-rust/pull/44. In my version, we preserve the property that `get_segments_for_output()` needs no heap allocation in the single-segment case, and we open the possibility of reusing heap space in the multi-segment case. The cost is that the caller needs to construct and pass in an `OutputCollector` struct. I think this is a totally reasonable trade off. What do you think?

danburkert:
Closing in preference of #45.

PR #44, opened 2015-05-31
OutputCollector allows get_segments_for_output() to take immutable self.

dwrensha:


danburkert:
I haven't compiled this example, but I think this allows a use after free:

```rust
let mut message = MallocMessageBuilder::new_default();
let mut collector = OutputCollector::new();

message.get_segments_for_output(&mut collector);
drop(message);
collector.get()
```

dwrensha:
I think that particular case is not a big problem because `get()` is a private method. However, your comment did make me realize that the reuse of `OutputCollector`s that I was imagining could lead to use-after-free if `clear()` were not called. I've push a new commit fixing that problem.

danburkert:
how about:

```rust
pub struct OutputCollector<'a> {
    segment0_for_output: &'a [Word],
    for_output: Vec<&'a [Word]>,
}

trait MessageBuilder {
    ...
    fn get_segments_for_output<'a>(&'a self) -> OutputCollector<'a>;
    ...
}
```

dwrensha:
Taking your last suggestion and running with it: https://github.com/dwrensha/capnproto-rust/pull/45

dwrensha:
Closing in favor of #45.

PR #48, opened 2015-08-21
Added SetPointerBuilder to Owned<'a>::Reader to maximise usefulness of Owned.

giacomocariello:
I've modified Owned<'a> so that Reader is constrainted with SetPointerBuilder. This way I can use an object X of type T::Builder with X.set_root(Y.get_root::<T::Reader>());

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -50,7 +50,7 @@ pub trait FromPointerReader<'a> {
 /// as a type parameter, e.g. for a generic container that owns a Cap'n Proto message
 /// of type `T: for<'a> capnp::traits::Owned<'a>`.
 pub trait Owned<'a> {
-    type Reader: FromPointerReader<'a>;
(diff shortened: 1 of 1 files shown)


dwrensha:
Thanks! There's a bit more work that needs to be done in [capnpc-rust](https://github.com/dwrensha/capnpc-rust) before this can be merged. In particular, we need to implement `SetPointerBuilder` for the `Client` types that are generated for [interfaces](https://capnproto.org/language.html#interfaces).

giacomocariello:
I see. I didn't notice that because I'm currently not using the rpc part.

The possibility of creating traits that implement generic reader-to-builder copy are crucial to my project, so I hope this feature gets official asap :-)

Thanks for your quick response.

dwrensha:
I've added the missing piece in https://github.com/dwrensha/capnpc-rust/commit/a8aa8a31f415132f2d34215ee6cb4e676f47a9e8.

PR #53, opened 2015-09-30
Allow iterating over lists of lists.

maurer:
Caveat: Getting an element of a list_list returns a Result<> which I had
to unwrap for IndexMove, so iterating over a malformed list may generate
a confusing error that can't be easily caught.

### DIFF
--- a/src/list_list.rs
+++ b/src/list_list.rs
@@ -21,7 +21,7 @@
 
 //! List of lists.
 
-use traits::{FromPointerReader, FromPointerBuilder};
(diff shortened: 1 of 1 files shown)


maurer:
Nevermind, realized when I typed out that caveat that I was doing something silly and fixed it. This version works as well (I've tested it with [Holmes](https://github.com/maurer/holmes/) by [iterating over a doubly nested list](https://github.com/maurer/holmes/blob/master/src/client.rs#L118-L125))

dwrensha:
Thanks!

PR #57, opened 2015-11-21
Miscellaneous cleanups

danburkert:
Surface level style cleanups, as well as removing a bunch of calls to transmute in `layout.rs`.


dwrensha:
Looks good. Thanks!

PR #66, opened 2016-09-12
aysnc serialization, compatible with futures-rs

dwrensha:
Using #42 as a starting point, adds `serialize::read_message_async()` and `serialize::write_message_async()`. These functions return a value that represents the in-progress I/O operation and can be driven to completion through the `poll()` method. 

Adds an optional dependency on `futures-rs` and implements `futures::Future` for the new in-progress I/O operations.

Modifies the existing `serialize::read_message()` and `serialize::write_message()` functions to use the new hand-rolled state machines under the hood. This appears to have negligible effect on performance.


dwrensha:
I've moved the new async functions into their own module, `serialize_async`, and I've reverted to using the existing serialize implementation in the blocking case.

dwrensha:
I've decided that this code belongs in its own crate. Hence: https://github.com/dwrensha/capnp-futures-rs

PR #69, opened 2016-12-02
Implement DoubleEndedIter for ListIter

Ralith:
This makes it convenient to walk backwards along a list in-place.

I haven't really tested this properly, but it might be a while before I get to that, so I'm submitting this now as the code's pretty simple anyway.


dwrensha:
Looks good to me. Thanks!

PR #72, opened 2017-05-27
remove dependency on Rc

vitiral:
Thre is no need for a `Rc` in this function (and this function is the only one that uses `Rc`), as the object is created in a function and the references are just passed to recursive function calls. The Rc gets you no safety and only eats performance.


vitiral:
@dwrensha  I think you got in the fix before me!

Issue #79, opened 2017-10-07
IntoIterator


crackcomm:
Is there any reason `IntoIterator` trait is not implemented anywhere in the code?

It would be useful at least for `capnp::primitive_list::Reader`.

dwrensha:
I think it would make sense to implement `IntoIterator` for the list types. Maybe you'd like to submit a pull request?

PR #80, opened 2017-10-13
IntoIterator trait for list readers

crackcomm:
Issue #79.

@dwrensha I must admit it is more of a copy and paste of your code than anything else :)


dwrensha:
Thanks!

Issue #75, opened 2017-07-24
Support rayon::iter::IntoParallelIterator in Reader and Builder as a feature


mmacedoeu:
As I see capnproto is all about non copy/ in-place casting so If at any point you need to deep copy all effort is lost. https://github.com/nikomatsakis/rayon is a parallel crate with easy parallel processing but to use it with capnproto without support you need to copy everything losing capnproto purpose.

dwrensha:
I agree that such support for parallel processing sounds useful! I don't think I'll have time in the near future to investigate exactly what would be involved; if you look into it and figure out some details, keep us posted!

Issue #50, opened 2015-09-07
Using Owned to use structs in-place instead of copying the data out


polyfractal:
I admit defeat :)  Continuing this question from IRC...I've been unable piece together how to appropriately use Owned.  I'm not familiar enough with capnp or Rust lifetimes to work out how the various pieces are used, and there aren't many examples in the three code repositories that use Owned.

My question on IRC: 
> any capnp tips to dealing with ownership? E.g. i want to serialize a timestamp, but to "rebuild" the timestamp on the other end, Chrono needs to take ownership of the u64 from capnp. I keep running into this with anything more advanced than simple primitives

And @dwrensha's response:

>capnp::traits::Owned might help: https://github.com/dwrensha/capnproto-rust/blob/f24ab9e95f1467ff023dd3b8a51495aa1637fa4d/src/traits.rs#L41-L56

>I think I eventually want to switch message::Reader and message::Builder to be parameterized by a `T where T : for<'a> Owned<'a>`.

>imagine that message::Reader is paramterized by foo::Owned. Then we would have `got_root<'a>(&'a self) -> foo::Reader<'a>`. The point is that the container doesn't need to worry about lifetimes; we can wait until a self-borrowing accessor method to commit to a lifetime.

>and `get_root<'a>(&'a self) -> <T as capnp::traits::Owned<'a>>::Reader`

Do you happen to have a small example handy? This seems like a critical part of using Capnp correctly.  I'd be happy to contribute back documentation once I figure out how the various pieces work together :)

dwrensha:
This is the kind of thing I had in mind:

```rust
    pub struct OwnedReader<T> where T : for<'a> ::capnp::traits::Owned<'a> {
        marker: ::std::marker::PhantomData<T>,
        message: ::capnp::message::Reader<::capnp::serialize::OwnedSegments>,
    }

    impl <T> OwnedReader<T> where T: for <'a> ::capnp::traits::Owned<'a> {
        pub fn get<'a> (&'a self) -> ::capnp::Result<<T as ::capnp::traits::Owned<'a>>::Reader >
        {
            self.message.get_root()
        }
    }
```

This lets you pass around an owned message without losing type information.

If the data you care about is not at the root of a message, then probably the easiest thing to do would be to copy the data into the root of a new message and then use the above pattern directly.

vks:
How would you implement `fn get<'a>(&'a self)` for an `OwnedBuilder`?

If I have

```rust
[...]

Issue #23, opened 2015-01-18
Capnproto messages are not Send


danburkert:
In trying to structure an app so that messages are read from a (TCP) stream, and then passed off to another thread for further processing, I'm running into issues because the message reader is not `Send`.  Is this intended, or perhaps fallout from the opt-in built-in changes?  For reference these are the errors I'm seeing when trying to move a message into a new thread:

```
error: the trait `core::marker::Send` is not implemented for the type `*const capnp::arena::SegmentReader`
error: the trait `core::marker::Send` is not implemented for the type `*const u8`
error: the trait `core::marker::Send` is not implemented for the type `*const capnp::layout::WirePointer`
```

Vaelden:
If you try to pass a Reader, I think this is intended behaviour, as a Reader only holds references to an object that actually owns the data: `capnp::serialize::OwnedSpaceMessageReader`.
So if you want to defer processing to another task, you have to send the object that actually owns the data, or you would have lifetime problems (holding references to an out-of-scope object).

Here is an example:
```rust
use capnp::serialize::OwnedSpaceMessageReader;
use std::sync::Arc;
fn test<T: std::io::Reader>(reader: &mut T, sender: &std::sync::mpsc::Sender<OwnedSpaceMessageReader>) {
    // Example 1: spawn a new thread
    let message_reader = capnp::serialize::new_reader(reader, capnp::message::DEFAULT_READER_OPTIONS).unwrap();
    std::thread::Thread::spawn(move || {
        let my_struct_reader = message_reader.get_root::<my_struct::Reader>();
        // do further processing in this task
    });

    // Example 2: send to an existing thead
    let message_reader_2 = capnp::serialize::new_reader(reader, capnp::message::DEFAULT_READER_OPTIONS).unwrap();
    sender.send(message_reader_2).unwrap();
}
```

dwrensha:
@Vaelden has it right. `MessageReader` is implemented for `SegmentArrayMessageReader` and `OwnedSpaceMessageReader`. The former cannot be `Send`, because it borrows the segments.

There's a similar situation for `MessageBuilder`, where `MallocMessageBuilder` is `Send` but `ScratchSpaceMallocMessageBuilder` is not.

[...]

PR #82, opened 2018-01-13
Add the OwnedReader as proposed in #50

bspeice:
Should also provide solutions for #23 and #75

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -25,7 +25,7 @@ use any_pointer;
 use private::arena::{BuilderArenaImpl, ReaderArenaImpl, BuilderArena, ReaderArena};
 use private::layout;
 use private::units::BYTES_PER_WORD;
-use traits::{FromPointerReader, FromPointerBuilder, SetPointerBuilder};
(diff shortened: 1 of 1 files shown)


dwrensha:
Thanks! I think we should also include the `get()` method as discussed in #50. Also, looks like you need to adjust some of the paths to get this to typecheck.

bspeice:
Fixed up, let me know if there are any other concerns.

dwrensha:
Hm. Now I'm thinking that `message::TypedReader` or similar might be a better name. The difference from `message::Reader` does not have to do with ownership actually -- it has to do with the extra type parameter `T`.

bspeice:
Ah, because we're no longer forcing the `OwnedSegments`. It's a trivial fix, just let me know.
What I do think may be more important - Should I add more to the docs/test cases to demonstrate moving across thread boundaries?

dwrensha:
OK, I'll merge this as-is and then fix up the names myself. Please feel free to submit new pull requests for docs and tests.

dwrensha:
renamed in https://github.com/capnproto/capnproto-rust/commit/f0efc35d7e9bd8f97ca4fdeb7c57fd7ea348e303

Issue #83, opened 2018-01-20
Build an owned version of a message


bspeice:
Alright, I think I finally know enough to properly phrase this question.

Right now it seems that the only way to create a `TypedReader` is to build a message using the standard pattern, serialize, deserialize (which then owns the segments), and stuff into the `TypedReader`.

Is there a way to build the message using `OwnedSegments` to start with? Or get at the `OwnedSegments` from the constructed message? If we're using `HeapAllocator`, the builder presumably owns the data and it could be safe to move from builder into `TypedReader`.

EDIT: I suspect it has something to do with `Builder::get_segments_for_output()`, but I'm not sure how to get the `OutputSegments` -> `OwnedSegments`

dwrensha:
Hm, interesting question! Maybe the way to go would be to implement `ReaderSegments` for `message::Builder`, like this:

```rust
impl <A> ReaderSegments for Builder<A> where A: Allocator {
    fn get_segment<'a>(&'a self, id: u32) -> Option<&'a [Word]> {
        self.get_segments_for_output().get(id as usize).map(|x| *x)
    }
}
```

Then you could create a `message::Reader` that wraps a `message::Builder`.

bspeice:
Sorry it's taken me so long to respond. That indeed works! The result is a bit strange (because you need to wrap the `Builder` inside a `Reader` to get at the `ReaderSegments`) but here's an example:

```rust
#[test]
fn test_typed_builder() {

    let mut builder = message::Builder::new_default();

    {
        let mut ts_root = builder.init_root::<timestamp::Builder>();

        ts_root.set_seconds(14);

        ts_root.set_subsecond_nanos(0);
    }

    let reader = message::Reader::new(builder, message::ReaderOptions::new());

    let owned_message: TypedReader<message::Builder<message::HeapAllocator>, timestamp::Owned> = TypedReader::new(reader);

    assert_eq!(owned_message.get().unwrap().get_seconds(), 14);
}
```

dwrensha:
Closed by #84.

PR #84, opened 2018-01-25
Convert (consume) Builder into TypedReader

bspeice:
Accidentally PR'd into my forked master, not original repo master.
Addresses capnproto/capnproto-rust#83

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -317,6 +317,18 @@ impl <A> Builder<A> where A: Allocator {
     pub fn get_segments_for_output<'a>(&'a self) -> OutputSegments<'a> {
         self.arena.get_segments_for_output()
     }
+    
+    pub fn into_typed<T: for<'a> Owned<'a>>(self) -> TypedReader<Builder<A>, T> {
+        let reader = Reader::new(self, ReaderOptions::new());
+
+        TypedReader::new(reader)
+    }
+}
+
+impl <A> ReaderSegments for Builder<A> where A: Allocator {
+    fn get_segment<'a>(&'a self, id: u32) -> Option<&'a [Word]> {
+        self.get_segments_for_output().get(id as usize).map(|x| *x)
+    }
 }
 
 #[derive(Debug)]
--- a/src/message.rs
+++ b/src/message.rs
@@ -21,6 +21,8 @@
 
 //! Untyped root container for a Cap'n Proto value.
 
+use std::convert::From;
+
 use any_pointer;
 use private::arena::{BuilderArenaImpl, ReaderArenaImpl, BuilderArena, ReaderArena};
 use private::layout;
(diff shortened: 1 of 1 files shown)


bspeice:
I don't know how we should handle the ReaderOptions - I.e. is it safe to just use the default, or should I expose the ability to pass in an external ReaderOptions?

dwrensha:
`ReaderOptions.traversal_limit_in_words` and `ReaderOptions.nesting_limit` exist to deal with untrusted data. When you have a `message::Builder`, you must have constructed the data yourself, so it's safe to set both of those options to `max_value()`. I think that's the right thing to do here.

Note that it also might make sense to convert a `message::Builder` to a `message::Reader` without going to a `message::TypedReader`, and it might make sense to have a `message::TypedBuilder` in the future. So I think it makes more sense to have separate methods: `message::Builder::into_reader()` and `message::Reader::into_typed()`.  We may also want to add an `message::TypedReader::into_inner()` message to allow recovering the reader.

bspeice:
Idiomatic Rust question then - Should it just be `TypedReader` [`From`](https://doc.rust-lang.org/std/convert/trait.From.html) both `message::Builder` and `message::Reader`? Saves one extra function call (`builder.into_reader().into_typed()`) and uses the Rust API's for it.

dwrensha:
I prefer adding methods with self-explanatory names like `into_reader()` and `into_typed()`. We could impl `From` on top of those methods, if we want that too.

bspeice:
All the above are done. Let me know if there's anything else, sorry about the wait on this one.

dwrensha:
Thanks!

dwrensha:
Hm... it occurs to me that the rest of the codebase uses methods named `as_reader()` instead of `into_reader()` for this kind of situation. I wonder if it makes sense to use that naming scheme here also, for the sake of consistency (even if `into_*()` is more idiomatic in the wider Rust community).

dwrensha:
`into_reader()` had an unnecessary type parameter: https://github.com/capnproto/capnproto-rust/commit/5bb2b41b704ea03ab3b4f08fab97e156b2a50da8

dwrensha:
After further reflection, I think the `into_` prefix is correct here. It signifies that the ownership is being transferred. In constrast, a prefix of `as_` usually signifies that something is being borrowed. It's true that many `as_reader` methods in capnproto-rust (e.g. [this one](https://github.com/capnproto/capnproto-rust/blob/5bb2b41b704ea03ab3b4f08fab97e156b2a50da8/src/struct_list.rs#L105)) technically take `self` by value, but their usual intended use case is more like borrowing, in particular when they are used in conjunction with a `borrow()` method. (See [this blog post](https://dwrensha.github.io/capnproto-rust/2014/12/27/custom-mutable-references.html) for more discussion of these methods.)

PR #86, opened 2018-02-08
Add write_message_segments/write_message_segments_to_words

polachok:
I'm not sure if I'm taking the right approach here, just wanted to gather feedback

https://github.com/capnproto/capnpc-rust/issues/45

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -311,3 +311,16 @@ impl <'a> ::std::ops::Deref for OutputSegments<'a> {
         }
     }
 }
+
+impl<'s> message::ReaderSegments for OutputSegments<'s> {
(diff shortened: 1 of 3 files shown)


dwrensha:
Thanks! I'll probably have a chance to take a look this weekend...

dwrensha:
This looks pretty good to me.

It's sad that we need the `get_reader_segment_count()` function. Maybe we should instead add a `len(&self) -> u32` method to `ReaderSegments`. At one point I think we may have been considering doing something fancy with `ReaderSegments` being sparsely populated, but now that seems to me unlikely to work out. A new `len()` method would need a default impl in order to be a non-breaking change.

polachok:
PR updated with len() method

dwrensha:
Thanks!

PR #99, opened 2018-08-02
Update fixes

khaledkbadr:
Remove deprecated code and fix some clippy warnings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,9 @@ addons:
       - gcc
       - g++
 install:
-  - curl -O https://capnproto.org/capnproto-c++-0.5.3.tar.gz
-  - tar zxf capnproto-c++-0.5.3.tar.gz
-  - cd capnproto-c++-0.5.3
+  - curl -O https://capnproto.org/capnproto-c++-0.6.1.tar.gz
+  - tar zxf capnproto-c++-0.6.1.tar.gz
+  - cd capnproto-c++-0.6.1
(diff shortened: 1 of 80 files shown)


dwrensha:
Thanks for looking at these things! I think a lot of them would require us to bump out minimum support rustc version from 1.15 to something newer. I've been planning to tie such a bump to a 0.9 release, which I hope will also update everything to Rust 2018 and use the new stdlib `Future` trait.

Maybe we should start a branch that you could submit such changes to?

khaledkbadr:
That would be a great idea if you want to keep 0.8 tied with rust 1.15

dwrensha:
I've decided that it would be good to go ahead with a 0.9 release (to be shipped sooner rather than later) that increases our minimum required rustc version and handles some long-planned API changes, without waiting on the Rust 2018 stuff.

I've pulled in your capnp upgrade: https://github.com/capnproto/capnproto-rust/commit/aa28b976949ebc60eac3662b8c2c5f1ed217083f

I've upped the rustc version to 1.26. Is that new enough for the some of the other stuff you want to do here?

khaledkbadr:
yup, that'd be enough until Rust 2018 is fully complete I think. I was planning to contribute but I wanted first to clear the deprecated stuff and fix the `clippy` warnings in order to use it as a style guide afterward.

dwrensha:
Closing, because this is out of date. Note that with #303 and associated diffs, we now have code formatting and linting as part of CI.

PR #109, opened 2018-10-06
Switch from try!() to ? in generated code to make it work in Rust 2018

bmisiak:
Rust 2018 removes the `try!()` macro and makes `try` a reserved keyword in accordance with [RFC 2388](https://github.com/rust-lang/rfcs/pull/2388). On the other hand, the `?` operator was introduced way back in Rust 1.13 and is now stable.

Without the change, rustc will refuse to compile the generated code starting with the 2018 edition:
```
expected expression, found reserved keyword `try`
```
 While this project itself is free to use `try!()` and not switch to Rust 2018, its output files - preferably - should be compatible with both editions.

https://github.com/rust-lang/rust/issues/31436


dwrensha:
Thanks!

PR #112, opened 2018-11-08
Address clippy warnings in autogenerated code

ysimonson:
This fixes most clippy warnings in capnpc autogenerated code, including:

* https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#needless_return
* https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#needless_lifetimes
* https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#cast_lossless
* https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#unit_arg

This does not semantically change the code that is generated, though the needless lifetimes fix might not be compatible with older versions of rustc (I seem to vaguely recall lifetime elison getting smarter at some point, so the explicit lifetimes may have been necessary to compile on older versions of rustc.)

For my schema, it reduces the warnings from several hundred to several dozen. The fix for the remaining warnings would require backwards-incompatible changes, so I will open a separate PR for that.


ysimonson:
Tests are passing locally when I run `cargo test`, but failing on TravisCI. Should I run something else to execute the entire test suite?

dwrensha:
error on Travis:
```
       self.builder.set_data_field::<i8>(2, i8::from(0u8));
     |                                            ^^^^^^^^ the trait `std::convert::From<u8>` is not implemented for `i8`
```

dwrensha:
Maybe try `cargo test --all`?

dwrensha:
Thanks!

PR #113, opened 2018-11-08
Address backwards-incompatible clippy warnings in autogenerated code

ysimonson:
In conjunction with #112, this addresses all of the clippy warnings I'm seeing for capnpc autogenerated code in my schema, by fixing for https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#wrong_self_convention

However, this is a breaking change, so it's in a separate PR. Calling code will have to change instances of `as_reader` to `into_reader`, and `from_server` to `into_client`.


ysimonson:
Note: this is not ready to merge yet. I'm seeing a lot of references to `as_reader` and `from_server` elsewhere in the codebase, and was hoping to get:

1) A thumbs up/down on whether this PR makes sense.
2) How I should proceed on addressing those references. Tests are already passing, but do I need to a command to re-generate some code, e.g. for `schema_capnp.rs`? Do I just need to find/replace all instances?

dwrensha:
I think this change makes sense. I've been reluctant to diverge from the names used in the analogous capnproto-c++ methods, but making the code idiomatic for Rust is probably more important than that concern.

The smoothest way to make the change would be to keep the old methods around and add deprecation annotations on them. Then people could gradually update. We would delete the old methods on the next version bump.

ysimonson:
OK, I added the methods back with deprecation notices. I've also updated the references to the old methods in tests and the readme.

dwrensha:
Hm... we also need to maintain consistency with the methods on the built-in types, of which there are many. For example, `text_list::Builder::as_reader()` would need to be deprecated in favor of `text_list::Builder::into_reader()`.

ysimonson:
That makes sense. I got all the methods I could find. I haven't updated the autogenerated code in `capnp/fuzz/fuzzers/test_capnp.rs`, `capnpc/src/schema_capnp.rs` - how would I go about doing that?

dwrensha:
Thanks!

dwrensha:
added some cleanup in  e20035304ef4785f542f67796c08b8b321e92af8

PR #115, opened 2018-11-18
More clippy linter improvements

ysimonson:
This includes ~~3~~ 2 changes:

* ~~#113 deprecated methods that produced clippy warnings, but they still produce warnings. This adds ignores for those lints, restricted to the deprecated methods only.~~ Could not apply this, due to an issue with stable/beta: https://github.com/rust-lang/rust/issues/54406
* #113 introduces a new clippy warning, because the `since` field in the `deprecated` annotation should be a semver string. This fixes that.
* Removed cases of unneeded unit expressions: https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#unused_unit

These are all in separate commits, so I can cherry-pick and open a new PR if you'd like only a subset of these fixes.

### DIFF
--- a/capnp/src/any_pointer.rs
+++ b/capnp/src/any_pointer.rs
@@ -157,7 +157,7 @@ impl <'a> Builder<'a> {
     }
 
     #[inline]
-    #[deprecated(since="capnpc-v0.9.2", note="use into_reader()")]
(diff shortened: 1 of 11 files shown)


dwrensha:
Thanks!

PR #119, opened 2018-12-01
Implemented support for Try trait to replace pry! for Promise

kabergstrom:
Got sick of the pry! macro so I implemented std::ops::Try for Promise to be able to use the `?` operator when using capnp-rpc. It can be enabled with the rpc_try feature.

I'm not sure about the crate::Error::unimplemented part, and not sure about whether the crate keyword itself requires a certain Rust version?


dwrensha:
> I'm not sure about the crate::Error::unimplemented part,

Yeah, it's weird that the `Try` trait requires conversions in both directions. I found some discussion of this topic: https://github.com/rust-lang/rust/issues/42327#issuecomment-312160563

In our case I think we only care about the `Result` -> `Promise` direction. Maybe we should leave it at this:

```rust
    fn into_result(mut self) -> Result<Self::Ok, Self::Error> {
        unimplemented!()
    }
```

kabergstrom:
Fixed this and tested it a little, seems to work fine for this use-case.

capnp-rpc/Cargo.toml
```diff
@@ -29,3 +29,6 @@ capnp = {version = "0.9.0", path = "../capnp",  features = ["rpc"]}
 # prevent 'cargo test' from trying to build the examples
 name = "dummy"
 path = "examples/dummy.rs"
+
+[features]
```
dwrensha: Do we gain anything by adding this feature here? It seems to me that anyone wanting to use it could more simply enable it directly on the capnp crate.

kabergstrom: No, you are right. I've pushed a fix to this.

dwrensha:
Thanks!

PR #120, opened 2019-01-02
Added output path option for build scripts

elniallo:
While it is not particularly rusty - being able to specify the out directory of the compiled code is useful, particularly in a test environment. If it is not set, it defaults to the environment set OUT_DIR


dwrensha:
Thanks!

Issue #93, opened 2018-04-06
Every message split into two packets


gavento:
Using capnp-rpc 0.8.3, every RPC message is split into two TCP packets: one 8 byte header with the segment_count and the 1st segment length, and the other with the actual first segment. While this does not break anything, sending two packets decreases performance.

I attach an example partial [packet capture](https://github.com/capnproto/capnproto-rust/files/1883549/rain-capnp-sample.zip) coming from the [rain](https://github.com/substantic/rain/) project (the connection creation is for example [here](https://github.com/substantic/rain/blob/14353f05365b576958b208f0ae6d2ed389f8be86/src/server/state.rs#L1159)).

PR #123, opened 2019-02-21
Add support for BufWriter wrapper

hukumka:
Enables usage of `std::io::BufWriter` to prevent message being
sent as separate packages.
Fixes https://github.com/capnproto/capnproto-rust/issues/93

Now `write_message` and `Transport` will flush (until success or error) after each message, 
allowing to write into buffered stream

### DIFF
--- a/capnp-futures/src/serialize.rs
+++ b/capnp-futures/src/serialize.rs
@@ -357,6 +357,7 @@ enum InnerWriteState {
         /// The byte offset into the next segment to write.
         idx: usize,
     },
+    Flush
 }
 
 impl InnerWriteState {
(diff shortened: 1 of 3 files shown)


dwrensha:
Thanks! Did you make any performance measurements? It would be nice to have some hard evidence that this buffering is an improvement.

hukumka:
I compared running time of rpc example calculator without and with buffer over loopback, and latter takes about 10% less time
violin

https://github.com/hukumka/capnp-calc-bench/tree/43d5ce7636dcaecfdd998a6a05c2324bbc92ef6d

But, I have not done yet a benchmark to see more exact effect

Also, here is a comparison of tcp packets for these cases
[tcpdump.zip](https://github.com/capnproto/capnproto-rust/files/2898244/tcpdump.zip)

dwrensha:
Thanks!

PR #126, opened 2019-04-11
Replace float transmutes with safe std functions.

solson:
This particular usecase has been safely abstracted in these `std` functions: [f32::to_bits](https://doc.rust-lang.org/std/primitive.f32.html#method.to_bits), [f32::from_bits](https://doc.rust-lang.org/std/primitive.f32.html#method.from_bits), [f64::to_bits](https://doc.rust-lang.org/std/primitive.f64.html#method.to_bits), [f64::from_bits](https://doc.rust-lang.org/std/primitive.f64.html#method.from_bits). One less `unsafe` to worry about in code review!

### DIFF
--- a/capnp/src/private/endian.rs
+++ b/capnp/src/private/endian.rs
@@ -84,18 +84,18 @@ endian_impl!(i64, to_le);
 
 impl Endian for f32 {
     fn get(&self) -> f32 {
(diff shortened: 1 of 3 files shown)


dwrensha:
Thanks!

PR #129, opened 2019-04-23
Problem: capnp::traits::Owned trait requirement (capnpc)

### LINKED ISSUE #128: Problem:  capnp::traits::Owned trait requirement (capnpc)
I've been trying to generate code for this (reduction of a more real-world scenario)

```
interface Base(T) {}

# Compiles fine
interface A(T) extends (Base(T)) {}

# Fails with a compile-time error (on Rust code)
interface A(T) extends (Base(Base(T))) {}
```

Where the error is:

```
660 | /   pub trait Server<T> : crate::test_capnp::base::Server<crate::test_capnp::base::Owned<T>> {
661 | |   }
    | |___^ the trait `for<'c> capnp::traits::Owned<'c>` is not implemented for `T`
    |
    = help: consider adding a `where for<'c> T: capnp::traits::Owned<'c>` bound
note: required by `test_capnp::base::Owned`
   --> 
    |
429 |   pub struct Owned<T> where T: for<'c> ::capnp::traits::Owned<'c>  {
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Any idea if this is a bug, or an expected behavior?

yrashk:
When trying to generate code for this:

```
interface Base(T) {}

interface A(T) extends (Base(T)) {}

interface A(T) extends (Base(Base(T))) {}
```

Compiling generated code fails with an error:

```
660 | /   pub trait Server<T> : crate::test_capnp::base::Server<crate::test_capnp::base::Owned<T>> {
661 | |   }
    | |___^ the trait `for<'c> capnp::traits::Owned<'c>` is not implemented for `T`
    |
    = help: consider adding a `where for<'c> T: capnp::traits::Owned<'c>` bound
note: required by `test_capnp::base::Owned`
   -->
    |
429 |   pub struct Owned<T> where T: for<'c> ::capnp::traits::Owned<'c>  {
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Solution: supplement required traits in the generated code

Closes #128

### DIFF
--- a/capnpc/Cargo.toml
+++ b/capnpc/Cargo.toml
@@ -30,3 +30,5 @@ path = "src/capnpc-rust2018.rs"
 [dependencies]
 capnp = { version = "0.9.5", path = "../capnp" }
 
+[build-dependencies]
+rustc_version = "0.2"
--- a/capnpc/build.rs
+++ b/capnpc/build.rs
@@ -1,0 +1,9 @@
+extern crate rustc_version;
+use rustc_version::{version, Version};
+
+fn main() {
+
+    if version().unwrap() >= Version::parse("1.30.0").unwrap() {
+        println!("cargo:rustc-cfg=rustc_at_least_1_30");
+    }
+}
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -1704,14 +1704,14 @@ fn generate_node(gen: &GeneratorContext,
                 Branch(vec!(
                     (if is_generic {
                         Branch(vec!(
-                            Line(format!("pub struct ToClient<U,{}> {{", params.params)),
+                            Line(format!("pub struct ToClient<U,{}> {} {{", params.params, params.where_clause)),
                             Indent(Box::new(Branch(vec!(
                                 Line("pub u: U,".to_string()),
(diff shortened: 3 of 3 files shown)


yrashk:
Looks like it just fails on 1.26 because I use [String#trim_start_matches](https://doc.rust-lang.org/std/string/struct.String.html#method.trim_start_matches) that is only available since 1.30

Should I fix this?

dwrensha:
It would be bad to increase the minimum required rustc version for a minor
change like this. So yes, we should find a way to avoid
trim_start_matches().

yrashk:
David, I will fix this shortly.

yrashk:
Ok, I've fixed this problem.

dwrensha:
Thanks! I've pulled in the first of your commits: https://github.com/capnproto/capnproto-rust/commit/dfb5e6cc85a0492a2993d9d67dad30729c255f5d
and then added my own resolution of the trim problem: https://github.com/capnproto/capnproto-rust/commit/2d13962a81ee530ea29542b9f03a7a78548f6f46

yrashk:
Excellent, thanks for a much better solution :)

PR #131, opened 2019-05-03
Treat UnexpectedEof as Disconnect

tecywiz121:
After the other end of a unix domain socket or a pipe disconnects, calls to `read_exact` will return an `UnexpectedEof` error.

I believe that this should be treated as an `ErrorKind::Disconnect` and not a generic `ErrorKind::Failure`.

### DIFF
--- a/capnp/src/lib.rs
+++ b/capnp/src/lib.rs
@@ -238,6 +238,7 @@ impl ::std::convert::From<::std::io::Error> for Error {
         use std::io;
         let kind = match err.kind() {
             io::ErrorKind::TimedOut => ErrorKind::Overloaded,
+            io::ErrorKind::UnexpectedEof |
             io::ErrorKind::BrokenPipe |
             io::ErrorKind::ConnectionRefused |
             io::ErrorKind::ConnectionReset |


dwrensha:
Thanks!

In the context of the RPC system, I agree that `ErrorKind::Disconnect` seems appropriate for this sort of error. But what if the (truncated) message is being read from disk? `ErrorKind::Disconnect` seems inappropriate in that case.

Maybe what we should really be doing is looking into wrapping errors more faithfully, as e.g. the [failure crate](https://github.com/rust-lang-nursery/failure) might allow us to do.

tecywiz121:
What would `ErrorKind::Disconnect` mean when reading from disk? That the file was deleted?

I've been using [`snafu`](https://crates.io/crates/snafu) lately, and been enjoying it, but you can avoid pulling in an external crate if all you want to do is wrap the error type: just add a variant with the error type and implement `std::error::Error`.

dwrensha:
I've added a `try_read_message()` function in 18c2e0cabaed06297c13f0a6504a806f9d53b914. The idea is that you can call this function in a loop and stop when it return None, which means EOF. Does that address your concern?

tecywiz121:
I'm no longer working on the project using this crate, but @JoelScarfone will probably have an opinion.

PR #135, opened 2019-06-08
use dyn Trait

tokahuke:
This line seems to be the only occurrence of a `Box<dyn Trait>`, `&dyn Trait` or `&mut dyn Trait` in code generation. Since `dyn` was introduced in Rust 2015, this does not break compatibility.

### DIFF
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -1632,7 +1632,7 @@ fn generate_node(gen: &GeneratorContext,
             mod_interior.push(
                 Branch(vec!(
                     Line(format!("impl {} ::capnp::capability::FromClientHook for Client{} {{", bracketed_params, bracketed_params)),
-                    Indent(Box::new(Line(format!("fn new(hook: Box<dyn ::capnp::private::capability::ClientHook>) -> Client{} {{", bracketed_params)))),
+                    Indent(Box::new(Line(format!("fn new(hook: Box<dyn (::capnp::private::capability::ClientHook)>) -> Client{} {{", bracketed_params)))),
                     Indent(Box::new(Indent(Box::new(Line(format!("Client {{ client: ::capnp::capability::Client::new(hook), {} }}", params.phantom_data)))))),
                     Indent(Box::new(Line("}".to_string()))),
                     Line("}".to_string()))));
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -1632,7 +1632,7 @@ fn generate_node(gen: &GeneratorContext,
             mod_interior.push(
                 Branch(vec!(
                     Line(format!("impl {} ::capnp::capability::FromClientHook for Client{} {{", bracketed_params, bracketed_params)),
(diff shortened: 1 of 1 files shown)


dwrensha:
Thanks!

Travis CI is seeing an error, and I see it too if I do `cargo test --all`.

Looks like `Box<dyn ::capnp::private::capability::ClientHook>` does not work in Rust 2015.
I think everything works if you change that to: `Box<dyn capnp::private::capability::ClientHook>`.

tokahuke:
Ok. I will amend that. I think I have got some spare time today.

tokahuke:
Surprisingly, Rust 2015 thinks `dyn` is a module (lol): https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&code=fn%20main()%20%7B%0A%20%20%20%20let%20_%20%3D%20Box%3A%3Anew(0..)%20as%20Box%3Cdyn%20%3A%3Astd%3A%3Aiter%3A%3AIterator%3CItem%20%3D%20i32%3E%3E%3B%0A%7D%0A

This can be solved with parentheses `Box<dyn (::capnp::private::capability::ClientHook)>`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&code=fn%20main()%20%7B%0A%20%20%20%20let%20_%20%3D%20Box%3A%3Anew(0..)%20as%20Box%3Cdyn%20(%3A%3Astd%3A%3Aiter%3A%3AIterator%3CItem%20%3D%20i32%3E)%3E%3B%0A%7D%0A

dwrensha:
ah, good find. using parens like that is probably better than omitting the
leading ::, as I suggested, because it does not introduce a risk of module
namespace collision. (I don’t fully understand the order of precedence for
module resolution; maybe there’s actually no risk.)

Does Rust 2018 warn on the parens? It might think they’re redundant.

tokahuke:
No. The Rust Playground example link compiles without warnings for 1.37. I remember the warning itself suggesting parentheses.

tokahuke:
Ooops! `dyn Trait` for Rust 1.26 is unstable. What do we do now?

dwrensha:
we could: 
 1. keep the generate code as-is, and add a `#[allow(bare_trait_objects)]` annotation,
 2. or we could bump our minimum-required rustc version from 1.26 to something newer.

If we do (2), we should also bump our crate version from `0.9.x` to `0.10.x`, as that would be a breaking change.

I'm inclined to do (2). There are some other changes that it'd be nice to slip in too with such a version bump.

dwrensha:
I should have some time this weekend. My plan is to consider merging this change and making the version bump then.

If you prefer some other option and would like to implement it, it probably makes sense to submit it as a separate pull request.

dwrensha:
In https://github.com/capnproto/capnproto-rust/commit/d4b793eb5d7e7ea47ced9bcb8515aa8a5b20ced1 I've bumped the minimum rustc version to 1.35 and bumped our version to 0.10.x.

tokahuke:
Great! Thanks.

PR #137, opened 2019-06-29
With output path

tokahuke:
So... I have found out that `CompilerCommand` fails when `OUT_DIR` is not set because of an `unwrap`. And OUT_DIR is not set for procedural macros, it seems. This pull request solves this issue introducing `CompilerCommand::with_output_path`, which never panics. It also gives a more detailed panic message on this behavior.

Should it be wise to also deprecate `CompilerCommand::output_path`, since it has become kinda redundant?

### DIFF
--- a/capnpc/src/lib.rs
+++ b/capnpc/src/lib.rs
@@ -77,10 +77,7 @@ pub enum RustEdition {
     Rust2018,
 }
 
-fn run_command(
-    mut command: ::std::process::Command,
-    path: &PathBuf,
-) -> ::capnp::Result<()> {
+fn run_command(mut command: ::std::process::Command, path: &PathBuf) -> ::capnp::Result<()> {
     let mut p = command.spawn()?;
     crate::codegen::generate_code(p.stdout.take().unwrap(), path.as_path())?;
     let exit_status = p.wait()?;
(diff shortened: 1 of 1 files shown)


dwrensha:
Hm... `with_output_path` makes the `output_path` option no longer follow the [Builder pattern](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html).

I think there are some simpler options:

  1. `CompilerCommand` could have `output_path: Option<PathBuf>` that starts as `None`. If it's still `None` on `run()`, then we panic (with a better error message than the current one).
  2. `CompilerCommand` initializes `output_path` to be `.` if the `OUT_DIR` environment variable is not set.

What do you think?

tokahuke:
Aggreed. I will refactor that in after work today.

tokahuke:
Ok. So, I went with the the `Option` approach and threw an error instead of panicking (not that actionable, though...). Messing with environment variables can be dangerous (mutable global state). Also, I wa not sure about the `.` default. It might be surprising...

dwrensha:
Looks good to me!

Could you please squash your changes into a single commit? Currently this pull request includes some commits (https://github.com/capnproto/capnproto-rust/pull/137/commits/eaa484b3145018a09c931e79b614436a51bee7a7, https://github.com/capnproto/capnproto-rust/pull/137/commits/8068d54bb4c1ee7ce6f21a6f91e73df8c13b2b7f) that have nothing to do with the intended change.

dwrensha:
Thanks!

PR #141, opened 2019-08-05
Fix #140 by exporting Disconnector struct from capnp_rpc

### LINKED ISSUE #140: Question: should Disconnector be part of capnp_rpc's exports?
I have a function `get_client` which returns a bootstrapped RPC client. I'd also like to return a `capnp_rpc::rpc::Disconnector`, so that the caller of `get_clinet` can run it when required. However, [`Disconnector`](https://github.com/capnproto/capnproto-rust/blob/master/capnp-rpc/src/rpc.rs#L1528) is not exported from the crate at the moment. Is this an oversight? Or am I using it wrong ?

ckoparkar:


dwrensha:
Looks good to me.
Thanks!

PR #142, opened 2019-08-09
Ignore dead code for generated source

AdminXVII:
Make sure not to pollute cargo's output with lots of dead_code warnings.


dwrensha:
Adding a top-level attribute like this makes it impossible to use `include!()` to pull in the generated code. See https://github.com/rust-lang/rust/issues/18810 and https://github.com/rust-lang/rfcs/issues/752.

I recommend manually inserting the `#[allow(dead_code)]` in the parent file of the generated code.

AdminXVII:
Yeah, that's what I did in the end. Thanks for the quick reply!

PR #150, opened 2019-09-27
add read_flat_message()

### LINKED ISSUE #149: read_message_from_words() should enable easy access to unconsumed words
[`read_message_from_words()`](https://github.com/capnproto/capnproto-rust/blob/81a1aed9de048e4c4d3f7f4d3fdd0b9d5ce8c678/capnp/src/serialize.rs#L53) permits the passed-in buffer to be longer than the message that it holds. (13632f825b1f65ecfa0b9b26602117c06474bf43.) However, if there is another message that starts after the first message, there is no easy way to get to it. We should write a better version of this function that exposes the number of words consumed, perhaps like this:

```rust
// (probably want a better name for this...)
struct FlatMessage<'a> {
   pub message: message::Reader<SliceSegments<'a>>,
   pub words_consumed: WordCount,
}

fn read_flat_message(slice: &'a [Word],
                     options: message::ReaderOptions)
                      -> Result<FlatMessage>;
```

dwrensha:
This is an enhanced version of read_message_from_words() that
makes it easy to know how many words were consumed in the
message.

Fixes #149.

### DIFF
--- a/capnp/src/serialize.rs
+++ b/capnp/src/serialize.rs
@@ -26,6 +26,7 @@ use std::convert::TryInto;
 use std::io::{Read, Write};
 
 use crate::message;
+use crate::private::units::BYTES_PER_WORD;
 use crate::{Error, Result, Word};
 
 /// Segments read from a single flat slice of words.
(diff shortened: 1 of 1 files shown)


dwrensha:
@bspeice does this look good to you?

bspeice:
I'm certainly not against it, but I'd throw out an alternative first: similar to [`read_segment_table`](https://github.com/capnproto/capnproto-rust/blob/master/capnp/src/serialize.rs#L103), make the `slice` pointer mutable and increment it by the amount of data read. This is already a low-level method, so working with the pointer isn't out of the question. And from an ergonomics perspective, it would allow you to keep calling the same method to read messages like a stream.

That said, modifying pointers as part of the public API feels kinda sketchy, I don't have a strong preference.

dwrensha:
I like that idea. I've updated this pull request to implement it.

I'm still debating the naming. It seems like this new method ought to also be called `read_message_from_words()`, but Rust does not allow overloading.

dwrensha:
Renamed it to `read_message_from_flat_slice()`.

Hm... I wonder whether we'll want to deprecate `read_message_from_words()` in favor of this new method. (It should be straightforward to transform all callsites of the former to use the latter.)

bspeice:
Deprecation makes sense; as your comment mentioned, there's not a good reason to use `read_message_from_words()` when there's another method that doesn't consume the slice.

PR #152, opened 2019-10-12
Replace leaf with juice on README

pheki:
Leaf has been unmaintained for a while (see https://medium.com/@mjhirn/tensorflow-wins-89b78b29aafb) and has been continued under https://github.com/spearow/juice


dwrensha:
Thanks!

PR #153, opened 2019-10-24
Include @generated in output file

Timmmm:
This means that code review tools know not to show the file by default. Rust does this for `cargo.lock` for example.


dwrensha:
I found some context on the cargo.lock version of this: https://github.com/rust-lang/cargo/issues/6180

dwrensha:
Thanks!

Timmmm:
Thanks for merging quickly! :-D

PR #154, opened 2019-10-24
Don't write output files if their content is unchanged

Timmmm:
Why would you want to do this? The issue is if you are writing the
output to your source directory, rather than your build directory,
which is a perfectly reasonable thing to do for various reasons.

In that case *always* writing the output causes issues in at least
two places so far:

1. If you have the rust-analyzer extension and tell it to watch the
   code for changes, it will get stuck in a loop. I suspect
   `cargo watch` will suffer from that too.
2. If you try to cross-compile using `cross` it will fail because it
   mounts the source code as a read-only filesystem.

I made a similar pull request for Prost here: https://github.com/danburkert/prost/pull/232

### DIFF
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -1874,6 +1874,15 @@ pub fn generate_code<T>(mut inp: T, out_dir: &::std::path::Path) -> ::capnp::Res
 
         let text = stringify(&lines);
 
+        let previous_text = ::std::fs::read(&filepath);
+        if previous_text.is_ok() && previous_text.unwrap() == text.as_bytes() {
+            // File is unchanged. Do not write it so that builds with the
+            // output as part of the source work in read-only filesystems
+            // and so timestamp-based build systems and watchers do not get
+            // confused.
+            continue;
+        }
+
         // It would be simpler to use the ? operator instead of a pattern match, but then the error message
         // would not include `filepath`.
         match ::std::fs::File::create(&filepath) {


dwrensha:
Looks good to me!

dwrensha:
Hm... I'm having second thoughts about this, as it could potentially double the memory usage of `capnpc-rust`.  

> writing the output to your source directory, rather than your build directory, which is a perfectly reasonable thing to do for various reasons.

Can you describe your use case a bit more in depth? If you're putting the generated files in your source directory, then I would expect that you would be OK with needing to manually trigger regeneration of them. Why aren't you putting these files in a build directory?

Timmmm:
Thanks for merging this!

> Why aren't you putting these files in a build directory?

* Code completion doesn't work (see [this issue](https://github.com/rust-analyzer/rust-analyzer/issues/1964)). Admittedly this is a temporary problem that will hopefully be fixed one day, but for now it makes a big difference.
* It's much easier to find and read the code!
* It makes code review easier - reviewers don't have to manually run the build and diff the build directories to see the changes to the generated files.

> it could potentially double the memory usage

Surely it will only increase memory usage by the size of the largest file it writes, which isn't going to be more than a few hundred kilobytes at most? If you're really bothered about it you could change it to a streaming read, rather than reading the whole file in at once, but frankly I think that is an insane thing to worry about. You know how much memory the rust compiler uses right? :-)

Issue #136, opened 2019-06-21
Amenities and ergonomics


tokahuke:
So... I have been working with Cap'n Proto in Rust for some time by now and although I quite happy with it, I find myself writing a lot of boilerplate. How viable would it be for we to create abstractions based on Rust's macro system (thinking Serde, Rokcet, etc...) to make coding experience more fun?

Is this even the scope of any of the current crates? 

I could help a bit on that.

realcr:
I am working with `capnproto-rust` for a pretty large project, and I also found using `capnproto-rust` directly as it is somewhat difficult. I find that I always have to define the data at least three times:

1. capnp file
2. Rust structs
3. Code that converts capnp generated Rust structures to my Rust structs.

I realize that working like this introduces performance penalty (One of the major selling points of capnp is having no encoding/decoding step). However, If I want to enjoy the no-encoding feature I will have to surrender to capnp and let it pollute my whole codebase with lifetimes and `.reborrow()`-s. Sprinkle some async Futures into the mix and you get a real party.

I am willing to give up on performance in favour of clear and easy to maintain code. (At the same time I realize that other people might not want to).

To be more specific about what my problem is, I want to introduce an example from my code base:

1. capnp structure:
```capnp
struct FriendsRoute {
        publicKeys @0: List(PublicKey);
        # A list of public keys
}

```

2. Rust structure
```rust
pub struct FriendsRoute {
    pub public_keys: Vec<PublicKey>,
}
```

3. Middle layer code, converting between capnp Rust structs and my Rust structs:

```rust
pub fn ser_friends_route(
    friends_route: &FriendsRoute,
    friends_route_builder: &mut funder_capnp::friends_route::Builder,
) {
    let public_keys_len = usize_to_u32(friends_route.public_keys.len()).unwrap();
    let mut public_keys_builder = friends_route_builder
        .reborrow()
        .init_public_keys(public_keys_len);

    for (index, public_key) in friends_route.public_keys.iter().enumerate() {
        let mut public_key_builder = public_keys_builder
            .reborrow()
            .get(usize_to_u32(index).unwrap());
        write_public_key(public_key, &mut public_key_builder);
    }
}
```

My plan is to have something like this:

```rust
#[capnp_conv(funder_capnp::friends_route)]
pub struct FriendsRoute {
    pub public_keys: Vec<PublicKey>,
}
```

[...]

PR #157, opened 2020-01-15
capnp-conv: proc macro ergonomics for capnproto-rust

realcr:
This PR adds a new crate called `capnp-conv`, allowing for automatic proc macro based code generation for serialization and deserialization.
Should solve issue https://github.com/capnproto/capnproto-rust/issues/136


dwrensha:
Thanks for the pull request!

I'm not going to have time to look at this immediately, partially because I'm focusing on [sorting out or alignment story](https://dwrensha.github.io/capnproto-rust/2020/01/11/unaligned-memory-access.html), but I hope to have a look before too long.

zimond:
@dwrensha is this happening anytime soon?

dwrensha:
It looks like this only works with `serialize_packed`.  Is there a way to generalize this so it would also work for the non-packed case?

capnp-conv/tests/capnp/test.capnp
```diff
@@ -0,0 +1,109 @@
...
+    externalUnion @17: TestUnion;
+    listUnion @18: ListUnion;
+}
+
+struct FloatStruct {
+    myFloat32 @0: Float32;
+    myFloat64 @1: Float64;
+}
+
+struct GenericStruct {
```
dwrensha: I gather that this is for testing a struct that generic on the Rust side. How does this library handle structs that are generic (i.e. have type parameters) on the capnp side? Like this one: https://github.com/capnproto/capnproto-rust/blob/4393eb558b4c4e855b04ab888a5adbc99a0eab0e/capnpc/test/test.capnp#L738-L744

capnp-conv/tests/capnp/test.capnp
```diff
@@ -0,0 +1,109 @@
+@0xc90daeac68e62b2a;
+
+struct TestStructInner {
+        innerU8 @0: UInt8;
+}
+
+struct TestUnion {
```
dwrensha: Can this library handle structs that have some union fields and some non-union fields? E.g.

```capnp
struct Foo {
   a @0 :UInt64;
   b @1 :Text;
   union {
      c @2 :Data;
      d @3 :UInt32;
  }
}
```

capnp-conv/tests/derive.rs
```diff
@@ -0,0 +1,217 @@
...
+fn capnp_serialize_floats() {
+    let float_struct = FloatStruct {
+        my_float32: -0.5f32,
+        my_float64: 0.5f64,
+    };
+
+    let data = float_struct.to_capnp_bytes();
+    let float_struct2 = FloatStruct::from_capnp_bytes(&data).unwrap();
+
+    // Sloppily check that the floats are close enough (We can't compare them directly, as they
```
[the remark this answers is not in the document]
realcr: Comparing floats is what makes me wake up in fear in the middle of the night. I am still new to understanding how it works in Rust, if you feel comfortable with this comparison I'm cool with it too.

capnp-conv/tests/capnp/test.capnp
realcr: capnp-conv doesn't handle the case of generic structs on the capnp side. I admit I am not really sure what would be the expected behaviour. This will require some thinking.

capnp-conv/tests/capnp/test.capnp
realcr: I am almost sure `capnp-conv` doesn't support this. What would be the Rust equivalent of this construct?

realcr:
@dwrensha : Thanks for the review!
1. I am not familiar with the packed vs non-packed case. Where can I get more information about it?
2. In case it was not obvious, you should feel free to make any changes to this piece of code (:

kscz:
>     1. I am not familiar with the packed vs non-packed case. Where can I get more information about it?

Hi @realcr - I know I wasn't who you asked, but since it appears no one is answering your question, I figured I would give it a shot!

One of the big things which makes Cap'n Proto unique compared to many other serialization formats which came before it, is that the in-memory representation is identical to the on-the-wire representation.  This means that "serializing" can, for all intents and purposes, simply be a memory-copy into the output buffer!  The downside is that this format means that details like padding bytes and unset optional fields end up being transmitted over the wire, which wastes bandwidth.

To mitigate this bandwidth issue, instead of having a "free" serialization step, you can use the "Packed" representation which basically tries to compress away as many zeros as possible.

For reading about either of these representations - https://capnproto.org/index.html talks about the "free" encoding step, and https://capnproto.org/encoding.html#packing talks about the "packed" representation.

opeik:
Hey there, could I be of any help on this issue? I'd appreciate some macro magic to reduce boilerplate.

ajosecueto:
any news?

lu-zero:
what is blocking this?

praveenperera:
Any updates?

p--b:
For those wanting to use this, NB that as an alternative, @aikilant has published https://github.com/aikalant/capnp_conv, which is available on crates.io as [`capnp_conv`](https://crates.io/crates/capnp_conv).

rupert648:
wanted to bump this again, this would be great to have in the core crate. Currently living with and maintaining a 800+ line serialisation file which is mostly similar boilerplate. Tests are also necessary to ensure that I don't miss out some serialisation call on a field. This would solve both those problems

PR #171, opened 2020-04-26
Remove static lifetime requirement

jannickj:
This PR attempts to remove the static lifetime requirement placed on futures.

I understand that I could be missing something given that this library is about providing bindings for a c++ library. That said looking at it, it seems to be correct.

I should note that this is a breaking change for the `ReadStream` struct, so a version bump will be required.


dwrensha:
Thanks! This looks good to me.

Because it's a breaking change, I think I'm going to hold off on merging this until I can batch it up with some other breaking changes. (Following our current scheme, bumping capnp-rpc to 0.13 would require bumping all of the capnp crates to 0.13.)

dwrensha:
Rebased and pushed to master: 57e7a28eea27a39d6470f5a882b9223d37d4e52a.

PR #179, opened 2020-05-14
giving reference to `init_x` methods

b00f:
Giving reference to init methods will help users to use initialize a parameter without re-borrowing it.

### DIFF
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -814,7 +814,7 @@ fn generate_setter(gen: &GeneratorContext, discriminant_offset: u32,
         Some(builder_type) => {
             result.push(Line("#[inline]".to_string()));
             let args = initter_params.join(", ");
-            result.push(Line(format!("pub fn init_{}(self, {}) -> {} {{",
+            result.push(Line(format!("pub fn init_{}(&self, {}) -> {} {{",
                                      styled_name, args, builder_type)));
             result.push(Indent(Box::new(Branch(initter_interior))));
             result.push(Line("}".to_string()));


b00f:
This 
https://github.com/capnproto/capnproto-rust/blob/23ee18c5844cd4ab39dfb8aea05f757cad79d609/capnpc/test/test_util.rs#L57
Can be changed to
```rust
let mut sub_sub_builder = sub_builder.init_struct_field();
```

dwrensha:
The `init_x` and `get_x` methods take `self` by value because `foo::Builder<'a>` is meant to act like a `&'mut Foo`. Please see this blog post: https://dwrensha.github.io/capnproto-rust/2014/12/27/custom-mutable-references.html

In particular, if `init_x` took `self` by reference, we not be able to write functions like this:
```rust
fn init_and_return_slice<'a>(foo : foo::Builder<'a>) -> &'a mut [u8] {
    foo.init_blob(100).slice_mut(5, 10)
}
```

b00f:
Oh, As I can see purposely  `init_x` method takes a by-value self parameter. I thought it might be a mistake. However I have some critics about that:

1-  `init_x` method is more similar to `set_x` rather than `get_x`. Both mutate `x`. They should have same parameter: `&mut self`

2- The get methods inside builder is kind of anti pattern. It moves the resource which is not a good practice.

3- As a user of this library (not the author) I prefer to not use have methods  like `init_and_return_slice` which moves foo. 
I'd rather prefer to have something like:
```rust
fn build_foo(foo : foo::Builder<'a>, x: f32, blob: &[u8])  {
    foo.set_x(x);
    foo.set_blob(blob);
}
```

or 

```rust
fn build_foo(foo : foo::Builder<'a>, x: f32, blob: &mut [u8])  {
    blob[0] = 42;
    foo.set_x(x);
    foo.set_blob(blob);
}
```

PR #183, opened 2020-05-23
remove assert! statements in CompilerCommand

Bromeon:
Remove asserts to allow for chaining/re-configuring the builder, as requested in https://github.com/capnproto/capnproto-rust/pull/181#issuecomment-632879591


dwrensha:
Thanks!

PR #184, opened 2020-05-23
add support for no-std

dwrensha:
Adds a "std" feature, enabled by default. Supports IO via a new module `capnp::io`, which defines custom `Read` and `Write` traits.

This is a breaking change from version 0.12, so we will need to bump the version before landing and publishing.


ProfFan:
Wow this is AWESOME! I think it's better to have a CI entry testing for `no_std` though as it is very easy to break `no_std` with "innocent" commits...

dwrensha:
Good point, @ProfFan. I've added some lines to the Travis config to exercise the no-std option, and now it's failing on stable and 1.40, as expected, because of the `async` blocks.

PR #188, opened 2020-06-19
fix read_segment_table read bug

gjcarneiro:
I was getting a corrupted read error.  I tracked the error down to this read_segment_table.

It has a loop that seems to suggest it was meant to repeatedly read even if the Read() call returns less than 8 bytes.  But if this happens, it reads again into the start of the buffer, instead of reading into the correct position.

I'm still trying to learn Rust, so please excuse me if this is not the correct fix.  But it does solve my case (with added prints, I found that Read() returned 7 instead o 8 after a few hundred messages).


dwrensha:
Good catch! Looks like I introduced this bug in 18c2e0cabaed06297c13f0a6504a806f9d53b914.

PR #199, opened 2020-08-16
Reduced number of dependencies pulled from futures

kabergstrom:
`futures` is kind of a heavy crate, and this PR reduces the number of sub-crates being included by capnp-rpc/capnp-futures by only pulling exactly the sub-crates that are used (futures_core, futures_channel, futures_io and futures_util).

Unfortunately, it seems my rustfmt didn't agree with yours - do you have any idea why?

### DIFF
--- a/capnp-futures/Cargo.toml
+++ b/capnp-futures/Cargo.toml
@@ -13,16 +13,22 @@ keywords = ["async"]
 
 [dependencies]
 capnp = { version = "0.13.0", path = "../capnp" }
+futures-core = { version = "0.3.0", default-features = false }
 
-[dependencies.futures]
+[dependencies.futures-io]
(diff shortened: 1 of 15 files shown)


dwrensha:
Good idea!

[Travis](https://travis-ci.org/github/capnproto/capnproto-rust/jobs/718449056) indicates that `cargo test -p capnp-futures` is failing. 

> Unfortunately, it seems my rustfmt didn't agree with yours - do you have any idea why?

Heh, I don't actually use rustfmt. 

Would it be possible to avoid the formatting changes in this pull request? It looks like a bunch of the lines that are changing are only formatting changes.

dwrensha:
This has been done in #527.

Issue #214, opened 2020-12-27
File capnp-rpc/schema/rpc-twoparty.capnp missing


Ekleog:
Hello!

I'm in the process of investigating my options for making rust code communicate with other rust code that runs inside a wasmtime sandbox, and capnp entered my radar.

However, in the process of investigating what would thus need to be done (it looks to me like I'll have to write an impl of VatNetwork, as I'd have to run over function calls which aren't AsyncRead/Write), I've come across https://github.com/capnproto/capnproto-rust/blob/master/capnp-rpc/src/rpc_twoparty_capnp.rs that references `capnp-rpc/schema/rpc-twoparty.capnp`

Yet this file is nowhere to be found. Would you happen to maybe have it locally and just have forgotten to check it in?

Sorry for bothering you and thank you for capnproto-rust!
  Leo

dwrensha:
That file got deleted for #163.
The canonical version of it is upstream: https://github.com/capnproto/capnproto/blob/348174a61642b094361fc6c98044933de3963ca2/c%2B%2B/src/capnp/rpc-twoparty.capnp

Ekleog:
Makes sense, thank you! I've just submitted https://github.com/capnproto/capnproto-rust/pull/215 so other people wouldn't ask the same question.

(BTW, do you have any form of chat where you'd hang out? I'm probably going to have quite a few other questions as I learn to use capnp, though I definitely can also spend more time reading the code and just opening a github issue if I find no way of doing what I'm trying to do)

dwrensha:
Thanks! I decided to just include the schema files in the repo instead: 869c3ff9f5c58461d90a6b1c94bbb8763a607c8c

dwrensha:
I do hang out on the Rust community Discord: https://bit.ly/rust-community

Ekleog:
Great, thank you!

PR #215, opened 2020-12-27
Fix source reference for capnproto-rust/capnp-rpc/src/rpc_twoparty_capnp.rs

Ekleog:
Fixes https://github.com/capnproto/capnproto-rust/issues/214


dwrensha:
Fix by 869c3ff9f5c58461d90a6b1c94bbb8763a607c8c instead.

PR #216, opened 2020-12-30
Fix broken link syntax in documentation

lassipulkkinen:


dwrensha:
Thanks! At one point rustdoc's markdown engine could handle line breaks like that. Apparently that's no longer the case!

PR #238, opened 2021-11-06
Implementing capnp::message::TypedBuilder

tomkris:
# Motivation
capnproto-rust already implements type `capnp::message::TypedReader` which is an owned capnp message which provides a strongly typed API for reading message content.

`TypedBuilder` provides same type of the functionality but for a message building case - it's a builder which owns content of the message being build and provides strongly typed API for building the message.

The main use case for this is to let developer define API which accepts a moved capnp message but wants to enforce it to be of a specific capnp type instead of accepting just a raw builder which may have some arbitrary content inside it.

# Implementation considerations
I tried to keep feature parity with regular `Builder` and closely replicate implementation details of the `TypedReader`

**1. I decided to have argument `T` at a first position (versus last position in `TypedReader`).**

```
pub struct TypedBuilder<T, A = HeapAllocator>
```
vs
```
pub struct TypedReader<S, T>
```

Motivation: I believe in most of cases developers will not be using custom allocator and keeping `A` type at last position will let them not specifying it at all, keeping code cleaner, e.g.

```
let x: TypedBuilder<my_struct::Owned>;
```
vs
```
let x: TypedBuilder<HeapAllocator, my_struct::Owned>;
```
or
```
let x: TypedBuilder<_, my_struct::Owned>;
```

**2. I implemented `TypedBuidler::new_default()` which implements an empty builder which is using `HeapAllocator`.**

I believe in most of cases developer just want to create a new empty builder without use of custom allocator.

More complicated cases with custom allocator or creating `TypedBuilder` based on `UntypedBuilder` are supported through `TypedBuilder::new(message: Builder<A>)`

**3. I replicated method naming schema of the `Builder` (e.g. `get_root`, `init_root`) and not schema of `TypedReader` (e.g. `get`).**

I believe consistent naming between `Builder` and `TypedBuilder` will make type more ergonomic and more familiar to existing develoeprs.

# Testing
```
cargo test
(cd capnpc/test && cargo test)
(cd capnpc/test && cargo test)
```


dwrensha:
Looks good to me. Thanks!

capnp/src/message.rs
```diff
@@ -20,7 +20,54 @@
...
+//! ```
+//!
+//! [TypedReader] and [TypedBuilder] accept generic type parameter `T`. This parameter must be
+//! a corresponding `Owned` type which was auto-generated inside the corresponding module.
+//!
+//! For example, for auto-generated module `crate::test_data::simple_struct` you'd supply 
+//! `crate::test_data::simple_struct::Owned` type into [TypedReader]/[TypedBuilder]
+//!
+//! ```ignore
+//! include!(concat!(env!("OUT_DIR"), "/simple_struct_capnp.rs"));
```
dwrensha, line 49: Directly including the generated code like this is not guaranteed to work. The usual pattern is to include the code inside a `mod`: https://github.com/capnproto/capnproto-rust/commit/42a460e6615f0889f237912daa5e4b1627a65e7e

Just now I put up a proposal for a macro that makes this less verbose: https://github.com/capnproto/capnproto-rust/pull/573

Issue #239, opened 2021-11-15
A `to_slice` method for primitive types


comath:
I want to store points for a machine learning library in cap'n proto, and would like to access the dense data as a slice of floats. Cap'n proto stores everything as little endian, and I believe it should be possible on little endian systems to safely return the data as slices. Is it possible and would it be welcome? 

Something like this:
```
#[cfg(target_endian = "little")]
pub trait PrimitiveListToSlice<T: PrimitiveElement> {
    fn to_slice(&self) ->&[T];
}
```
I believe it only reads on contiguous data, so we don't need to worry about the message being split over a few buffers. If we do, we can return an `Option`, or a `Vec` of slices, but that's way more complicated than what I think we need to do to add this feature.

dwrensha:
We do have something along these lines in `capnp::raw::get_list_bytes()`:

https://github.com/capnproto/capnproto-rust/blob/9cf00528a6a92ba57c835fd467f5217d84e5707b/capnp/src/raw.rs#L54-L59

So I suppose the only missing piece would be the conversion from `&[u8]` to `&[f32]`.
My initial feeling is that you should write your own little function do that `transmute` on your end, but I could maybe be convinced that something like that belongs within the `capnp` crate.

comath:
I'm working on a evaluation library, and have added the recommended transmute in there. The transmute needs a check for little-endianness and a use of unsafe. It gets a little messy, and I'm going to need to do this a lot of places.

[Flatbuffers](https://docs.rs/flatbuffers/2.0.0/flatbuffers/struct.Vector.html#method.safe_slice) have a 'to_slice' method :grin:.

dwrensha:
closed by #240.
renamed to `as_slice()` in f77e49986df561b82de8219affda75b043f1afb9

PR #240, opened 2021-11-30
`to_slice` for little endian systems 

comath:
Adds a `to_slice` call for the `primitive_list` `Reader` and `Builder` structs on little endian systems. For the reader it produces a just a slice, and for the builder it produces a mutable slice. 

This is to resolve issue https://github.com/capnproto/capnproto-rust/issues/239

### DIFF
--- a/capnp/src/primitive_list.rs
+++ b/capnp/src/primitive_list.rs
@@ -25,7 +25,7 @@ use core::{marker};
 
 use crate::traits::{FromPointerReader, FromPointerBuilder, IndexMove, ListIter};
 use crate::private::layout::{ListReader, ListBuilder, PointerReader, PointerBuilder,
-                             PrimitiveElement};
+                             PrimitiveElement, data_bits_per_element};
 use crate::Result;
 
 #[derive(Clone, Copy)]
@@ -71,6 +71,15 @@ impl <'a, T: PrimitiveElement> Reader<'a, T> {
         assert!(index < self.len());
         PrimitiveElement::get(&self.reader, index)
     }
+
+    #[cfg(all(target_endian = "little"))]
+    pub fn to_slice(&self) -> &[T] {
+        let bytes = self.reader.into_raw_bytes();
+        unsafe {
+            use std::slice;
+            slice::from_raw_parts(bytes.as_ptr() as *mut T, (data_bits_per_element(T::element_size())/8) as usize)
+        }
+    }
 }
 
 impl <'a, T> crate::traits::IntoInternalListReader<'a> for Reader<'a, T> where T: PrimitiveElement {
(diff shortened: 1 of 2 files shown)


capnp/src/primitive_list.rs
```diff
@@ -71,6 +71,15 @@ impl <'a, T: PrimitiveElement> Reader<'a, T> {
         assert!(index < self.len());
         PrimitiveElement::get(&self.reader, index)
     }
+
+    #[cfg(all(target_endian = "little"))]
+    pub fn to_slice(&self) -> &[T] {
+        let bytes = self.reader.into_raw_bytes();
+        unsafe {
+            use std::slice;
+            slice::from_raw_parts(bytes.as_ptr() as *mut T, (data_bits_per_element(T::element_size())/8) as usize)
```
dwrensha, line 80: It's possible that `self.reader.get_element_size()` is larger than `T::element_size()`. See https://capnproto.org/language.html#evolving-your-protocol:
> A field of type `List(T)`, where `T` is a primitive type, blob, or list, may be changed to type `List(U)`, where `U` is a struct type whose `@0` field is of type T.

In such a case, this function will return bad data.

dwrensha: Also, the second argument to `slice::from_raw_parts()` needs to use `bytes.len()` somehow. 
(Looks like the `Builder` version of `to_slice()` got this correct.)

comath: I made it return an option, with something if the underlying data was still of the right number of bits or nothing if it's migrated.

capnp/src/primitive_list.rs
```diff
@@ -97,6 +111,15 @@ impl <'a, T> Builder<'a, T> where T: PrimitiveElement {
     pub fn set(&mut self, index: u32, value: T) {
         PrimitiveElement::set(&self.builder, index, value);
     }
+
+    #[cfg(all(target_endian = "little"))]
+    pub fn to_slice(&self) -> &mut [T] {
+        let bytes = self.builder.into_raw_bytes();
+        unsafe {
```
dwrensha, line 118: `self.builder.element_size != T::element_size()` is possible in this case too. (For example, we could have an `any_pointer::Builder` and initialize it as a struct list but then call `get_as()` as a primitive list.)

comath: That's a really weird use, but possible. :D 
I made this perform the same check and return an option as well.

dwrensha:
Added some fixups (f77e49986df561b82de8219affda75b043f1afb9, 70cfba18eceacef38c18c55fc01d9c93e2f77d4e) and merged.

Thanks!

Issue #241, opened 2021-12-13
Remove need for `reborrow()`


jyn514:
I looked at the 2014 blog post, which says:

> The `Builder` types of [capnproto-rust](https://github.com/dwrensha/capnproto-rust) also need to
provide an exclusivity guarantee.  Recall that if `Foo` is a struct defined in a Cap'n Proto schema,
then a `foo::Builder<'a>` provides access to a writable location in arena-allocated memory that
contains a `Foo` in [Cap'n Proto format](https://kentonv.github.io/capnproto/encoding.html).  To
protect access to that memory, a `foo::Builder<'a>` ought to behave as if it were a `&'a mut Foo`,
even though the `Foo` type cannot directly exist in Rust (because Cap'n Proto struct layout differs
from Rust struct layout).

The solution it comes up with is to use `reborrow()` everywhere. But that's not necessary, since it can just take `&mut self` in the accessors instead of taking an owned value; that both enforces exclusive access and allows reusing the builder later without needing a reborrow() call.

jyn514:
Note that this is a breaking change, since `let my_builder = Builder::new_default(); my_builder.init_x()` compiles if `init_x` takes `self`, but not if it takes `mut self` (the user has to update their code to use `let mut my_builder`).

dwrensha:
The problem with taking `&mut self` is that then you can't write functions that return mutable references to the interior of a message. For example, suppose we have these types:

```capnp
struct Foo {
  blob @0 : Data;
}
struct Bar {
  foo0 @0 : Foo;
  foo1 @1 : Foo;
}
```

and suppose that I want to write a function that takes a mutable reference to a `Bar` and returns a mutable reference to a `Foo`.

With the current setup I can write a function like this:
```rust
fn return_foo_mut<'a>(get_foo0: bool, mut bar: Bar::Builder<'a>) -> capnp::Result<Foo::Builder<'a>> {
   if get_foo0 { bar.get_foo0() } else { bar.get_foo1() }
}
```

How would I write an equivalent function after making the changes that you are proposing?

jyn514:
You'd change the function to take a mutable reference instead of an owned value:
```rust
fn return_foo_mut<'a>(get_foo0: bool, mut bar: &'a mut Bar::Builder<'_>) -> capnp::Result<Foo::Builder<'a>> {
   if get_foo0 { bar.get_foo0() } else { bar.get_foo1() }
}
```

dwrensha:
[...]

PR #242, opened 2021-12-13
Make `reborrow()` unnecessary in most cases

jyn514:
This is what the fix for https://github.com/capnproto/capnproto-rust/issues/241 would look like; happy to extend this to all the other uses of `reborrow()` if you like :) I just want to make sure you think this is a good idea before putting too much time into it.

- get() doesn't consume the data being read from, so reflect that in the type signature.
- `init_*()` can take `&mut self` instead of consuming `self`
- update lots of examples

### DIFF
--- a/capnp/src/any_pointer_list.rs
+++ b/capnp/src/any_pointer_list.rs
@@ -52,7 +52,7 @@ impl <'a> Reader<'a> {
         ListIter::new(self, l)
     }
 
-    pub fn get(self, index : u32) -> crate::any_pointer::Reader<'a> {
+    pub fn get(&self, index : u32) -> crate::any_pointer::Reader<'a> {
         assert!(index <  self.len());
         crate::any_pointer::Reader::new(self.reader.get_pointer_element(index))
     }
@@ -87,7 +87,7 @@ impl <'a> Builder<'a> {
         Reader { reader: self.builder.into_reader() }
     }
 
-    pub fn get(self, index : u32) -> crate::any_pointer::Builder<'a> {
+    pub fn get(&mut self, index : u32) -> crate::any_pointer::Builder<'a> {
         assert!(index <  self.len());
         crate::any_pointer::Builder::new(self.builder.get_pointer_element(index))
     }
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -908,7 +908,7 @@ fn generate_setter(gen: &GeneratorContext, discriminant_offset: u32,
         Some(builder_type) => {
             result.push(Line("#[inline]".to_string()));
             let args = initter_params.join(", ");
-            result.push(Line(format!("pub fn init_{}(self, {}) -> {} {{",
+            result.push(Line(format!("pub fn init_{}(&mut self, {}) -> {} {{",
                                      styled_name, args, builder_type)));
             result.push(Indent(Box::new(Branch(initter_interior))));
             result.push(Line("}".to_string()));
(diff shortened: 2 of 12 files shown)


capnpc/src/codegen.rs
```diff
@@ -1280,15 +1280,15 @@ fn generate_node(gen: &GeneratorContext,
...
+                            Line(format!("pub fn get_{}(&self) {} {{", styled_name, ty)),
                             Indent(Box::new(get)),
                             Line("}".to_string()))));
 
                     let (ty_b, get_b, _) = getter_text(gen, &field, false, true)?;
                     builder_members.push(
                         Branch(vec!(
                             Line("#[inline]".to_string()),
-                            Line(format!("pub fn get_{}(self) {} {{", styled_name, ty_b)),
+                            Line(format!("pub fn get_{}(&self) {} {{", styled_name, ty_b)),
```
dwrensha, line 1291: Needs to be `&mut self` to enforce non-aliasing of mutable references.

capnp/src/any_pointer_list.rs
```diff
@@ -87,7 +87,7 @@ impl <'a> Builder<'a> {
         Reader { reader: self.builder.into_reader() }
     }
 
-    pub fn get(self, index : u32) -> crate::any_pointer::Builder<'a> {
+    pub fn get(&self, index : u32) -> crate::any_pointer::Builder<'a> {
```
dwrensha, line 90: Needs to be `&mut self` to prevent aliasing of mutable references.
(Similar comment applies elsewhere in this diff.)

capnpc/src/codegen.rs
jyn514: Hmm, I don't quite follow - how is it possible to mutate through the object this returns?

dwrensha: Without this `mut`, we could have
```capnp
struct Foo {
  blob @0 : Data;
}
```

```rust
    fn return_aliased<'a>(foo: &'a foo::Builder<'_>) -> capnp::Result<(&'a mut [u8], &'a mut [u8])> {
        Ok((foo.get_blob()?, foo.get_blob()?))
    }
```

I.e. a function that returns two mutable slices pointing to the same memory.

dwrensha: Probably `capnp::private::StructBuilder` should *not* derive `Copy`: https://github.com/capnproto/capnproto-rust/blob/2832d59d197ff9c99603e659f02dad0bdd8dfdee/capnp/src/private/layout.rs#L2980-L2981

Then then methods would be forced to take a `&mut self` and it would be less confusing.

dwrensha: hm... actually I think that in order for safety to work out we need it to be `&'a mut self` (so that the lifetime is constrained to be the same as the lifetime of the returned `Builder<'a>`). And once we make that change, I think that unfortunately the examples we've been talking about fail to pass the borrow checker.

dwrensha: A concrete illustration of what can go wrong if `Builder::get` takes `&mut self` instead of `&'a mut self`:
```rust
    fn return_aliased<'a>(foo: &'a mut foo::Builder<'_>) -> capnp::Result<(&'a mut [u8], &'a mut [u8])> {
        Ok((foo.get_blob()?, foo.get_blob()?))
    }
```

That is, again we find that it's possible to have multiple mutable references to the same memory.

dwrensha:
Thanks for the pull request and the discussion. Closing, because I don't see a way to make this work.

PR #243, opened 2021-12-14
Adds `BufferSegments` for long lived mmap support

comath:
This adds a new struct that's a partner to the `SliceSegments` struct. It's meant to hold an owned buffer like a mmap: 
```rust
use memmap2::Mmap;
use std::fs::File;
use capnp;

// We know the loaded message is safe, so we're allowing unlimited reads.
let unlimited_reads = ReaderOptions {
    traversal_limit_in_words: None,
    nesting_limit: 64,
};

let file = File::open("practically_static_message.capnp")?;
let mmap = unsafe { Mmap::map(&file).unwrap() };
let segments = BufferSegments::new(value, unlimited_reads)?;

let reader = capnp::message::Reader::new(segments, unlimited_reads).into_typed();
```
This struct lets me work with mmaps with little headache and I've been using it for a few weeks now. Should help others too, like [#121](https://github.com/capnproto/capnproto-rust/issues/121). 

Reason: I'm using capnp for a k-nearest neighbors library where 95% of what I need to point to are essentially dense vectors of `&[f32]` and sparse vectors of `(&[f32],&[u32])` (more or less). Parsing the pointers to the point I need takes about as much time as the distance to an the distance between two 4096 dimension dense vectors (on my machine) and is dwarfed by the mmap lookup to a NVMEe raid0 array. With some additional parsing at load the prototype capnp implementation of the underlying point storage is about 2x faster than the old protobuf one and has much less code.

### DIFF
--- a/capnp/src/serialize.rs
+++ b/capnp/src/serialize.rs
@@ -32,6 +32,8 @@ use crate::message;
 use crate::private::units::BYTES_PER_WORD;
 use crate::{Error, Result};
 
+use std::ops::Deref;
+
 /// Segments read from a single flat slice of words.
 pub struct SliceSegments<'a> {
     words: &'a [u8],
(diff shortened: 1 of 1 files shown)


marmeladema:
Hello and thank you for this!
I was about to make a PR for a similar improvement. In my case, I modified the `SliceSegments` to store a `Cow<'a, [u8]>` but in the end the user facing APIs are pretty much the same.

capnp/src/serialize.rs
```diff
@@ -86,6 +88,66 @@ pub fn read_message_from_flat_slice<'a>(slice: &mut &'a [u8],
...
+        assert!(segment_table.total_words() * 8 <= buffer.len());
+        let segment_indices = segment_table.to_segment_indices();
+        Ok(BufferSegments {
+            buffer,
+            segment_table_bytes_len,
+            segment_indices,
+        })
+    }
+
+    pub fn to_buffer(self) -> T {
```
marmeladema, line 128: nit: usually methods that consume `self` to convert the object into something else are prefixed by `into_` rather than `to_`

capnp/src/serialize.rs
```diff
@@ -86,6 +88,66 @@ pub fn read_message_from_flat_slice<'a>(slice: &mut &'a [u8],
...
+        self.buffer
+    }
+}
+
+impl <T: Deref<Target = [u8]>> message::ReaderSegments for BufferSegments<T> {
+    fn get_segment<'b>(&'b self, id: u32) -> Option<&'b [u8]> {
+        if id < self.segment_indices.len() as u32 {
+            let (a, b) = self.segment_indices[id as usize];
+            Some(&self.buffer[self.segment_table_bytes_len..][(a * BYTES_PER_WORD)..(b * BYTES_PER_WORD)])
```
marmeladema, line 137: Instead of extracting a slice twice, maybe just `[(self.segment_table_bytes_len + a * BYTES_PER_WORD)..(self.segment_table_bytes_len + b * BYTES_PER_WORD)]` would be more efficient?

capnp/src/serialize.rs
```diff
@@ -86,6 +88,66 @@ pub fn read_message_from_flat_slice<'a>(slice: &mut &'a [u8],
     }
 }
 
+/// Segments read from a buffer, useful for long lived mmaps.
```
marmeladema, line 91: My use case is that I own a message (as a `Vec<u8>`) and I need to return a string extracted from the message. Because of the `'a` associated to `self` in https://docs.capnproto-rust.org/capnp/message/struct.Reader.html#method.get_root I need to keep the `message::Reader` alive and store it *somewhere*.
However the only place I can store it is in the struct that already contains the message buffer, but then I cannot read it using the `SliceSegments` because it would make it a self-referential struct. I have two solutions:
* Read the buffer back with https://docs.capnproto-rust.org/capnp/serialize/fn.read_message.html at the cost of basically duplicating the message which I want to avoid for performance reasons
* Or use the API you introduced to actually *store* my buffer in the `BufferSegments` which is stored in the `Reader` without any copy or heap allocation :+1: 

All of that to say that this `BufferSegments` is not only useful for long lived mmaps but actually in many other scenario as well.

comath: I use it for that too, I have an enum with a memmap2 `Mmap` and a `Vec<u8>` and implement `Deref` on that. Grabbing large objects with `rusoto_s3` for these long lived messages is most naturally done by reading to `Vec<u8>`.

dwrensha:
Thanks!

dwrensha:
added some touchups: https://github.com/capnproto/capnproto-rust/commit/267b8b8939fb63b1765741617b98d1029466c917

PR #247, opened 2022-01-02
Add raw schema output option to capnpc

### LINKED ISSUE #246: Using CodeGeneratorRequest directly
Is there any way to access the `CodeGeneratorRequest` object from what was generated in the `build.rs`? What I have in mind is a `proc_macro` that implements additional traits for generated Reader and Builder structs based on custom annotations in the `.capnp` file?

If there is no way for that, can I write a simple PR that adds a method to `capnpc::CompilerCommand` that is similar to [`file_descriptor_set_path`](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.file_descriptor_set_path) from `prost_build` to output a serialized `CodeGeneratorRequest` into the `OUT_DIR` so I can read it from my `proc_macro`? Or maybe some different way to access it like returning a `CodeGeneratorRequest` reader from a separate version of `CompilerCommand::run()` function?

freopen:
Closes #246

### DIFF
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -36,6 +36,7 @@ use self::FormattedText::{Indent, Line, Branch, BlankLine};
 pub struct CodeGenerationCommand {
     output_directory: PathBuf,
     default_parent_module: Vec<String>,
+    raw_code_generator_request_path: Option<PathBuf>,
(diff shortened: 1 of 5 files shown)


dwrensha:
Thanks!

PR #265, opened 2022-03-19
Mention "level 1" in the main readme

### LINKED ISSUE #264: Docs: which Cap'n Proto RPC level does this rust implementation satisfy?
Hi there,

I just recently learned about cap'n proto and this library. The [RPC docs page](https://capnproto.org/otherlang.html) for the protocol breaks the total feature set up into 4 "levels" which implementations can advertise in order to succinctly communicate to users how much of the specification it implements.

Which level does this implementation satisfy? And would it be appropriate to add this information to the README?

Thanks!
Oliver

OliverEvans96:
This might not be the optimal placement, but it allows prospective users to ctrl-f "level" when opening the repo. 

Closes #264.


dwrensha:
Hm. I don't think it makes sense to have this quite so prominent, above the core serialization features.

How about we put it in the "Crates" table instead? 

```
| [capnp-rpc](/capnp-rpc) | Object-capability remote procedure call system with ["level 1"](https://capnproto.org/rpc.html#protocol-features) features | [![crates.io](https://img.shields.io/crates/v/capnp-rpc.svg)](https://crates.io/crates/capnp-rpc) |
```

OliverEvans96:
Yeah, good point - that's a better spot. I updated my branch with your suggestion. Thanks for the feedback.

dwrensha:
Thanks!

PR #266, opened 2022-04-28
Lookup super interfaces recursively

zortax:
Currently, only one level of interface inheritance is supported. This PR adds a recursive lookup to support multiple levels.

### DIFF
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -1893,12 +1893,31 @@ fn generate_node(gen: &GeneratorContext,
             }
 
             let mut base_dispatch_arms = Vec::new();
+
             let server_base = {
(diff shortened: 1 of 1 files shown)


dwrensha:
What's an example of a schema that does not work before this change?

zortax:
```capnp
interface A {
    someMethod @1 ();
}

interface B extends(A) {}

interface C extends(B) {}
```

calling `someMethod` on a `C`-client causes a Method-not-implemented error, as the type ID of that call doesn't match the interface nor the direct super interface (there was no branch calling the correct ServerDispatch in the type id match statement being generated)

zortax:
@dwrensha any chance this could be merged soon?

dwrensha:
Thanks!

Issue #182, opened 2020-05-23
Distribution of the capnp tool for Rust projects


Bromeon:
In #181, a discussion about distributing the `capnp` schema generator tool (just called "tool" from now on) alongside Cap'n'Proto arised.

Status quo according to https://capnproto.org/install.html is that Windows provides binaries for download, whereas on Unix and OS X a package manager provides installations (managed by third parties). Additionally, on all platforms the tool can be built from source.

When developing a Rust application, it is common to use the crate `capnpc` in `build.rs` for schema generation. At the moment, this still requires the tool to be installed externally. This is generally acceptable, but can be a hindrance when aiming for a self-contained Rust project, e.g. for collaboration, continuous integration or deployment.

For distributing the tool, I see the following options:

### 1. Install separately or let the user select binaries
This is the current solution. From #181 onwards, it is possible for a Cap'n'Proto user to provide binaries for the different platforms himself and ship them along the project.

### 2. Build from source
`build.rs` would invoke CMake and build the tool from C++ source. I haven't tried this yet to see how complex the process would be.

It would require a C++ toolchain installed on the target platform however, which may not always be the case for Rust projects.

### 3. Ship precompiled binaries
The tool could do the same as [Prost, a Rust protobuf plugin](https://github.com/danburkert/prost/tree/master/prost-build/third-party/protobuf) and provide executables for the most common platforms. `capnpc` would then select the correct one depending on the platform.

A few points would need to be considered:
* how are the executables kept up-to-date? is it required to manually recompile for all platforms?
* how often is the tool updated?
* should the tool be part of the `capnpc` crate? a separate crate would have several advantages:
  * binaries can be kept at their own version (often lower than `capnpc` one)
  * a separate Git repo avoids bloating this one with binary diffs
  * release cycles of `capnpc` are not tied to the binary compilation, so there would be no delay because binaries are not built yet

### 4. RIIR
Just mentioning this because it's bound to come up sooner or later 😬 
Rewriting the tool in Rust comes at quite a cost and risk, and either will shift the problem to the C++ side, or require maintenance of two tools. I don't see this as a viable solution.

---

Obviously, besides convenience for Cap'n'Proto users, we should also try to minimize the extra maintenance needed on @dwrensha's side.

Maybe it's interesting to see how other users are using Cap'n'Proto, and what are current practices. It may turn out that this is not seen as a problem at all. Share your experiences!

zenhack:
[...]

PR #268, opened 2022-05-15
capnpc: add option to build `capnp` exe from src

rileylyman:
Related to #182.

This is one potential way we could implement option `2.)` while keeping in mind @zenhack's concern about pushing users away from their package manager. The idea is that we check for the existence of `capnp` at build time. If it does not exist, we throw an error letting the user know to either install it OR provide `feature = "build-capnp"` to the crate so that we can build it for them. 

I think that it is good to at least provide this option to the user since many may find it desirable to have a plug-and-play solution without the need for prerequisites when building their crate.

You may find adding the `capnproto` submodule undesirable, in which case we could potentially clone the repo at build time.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "capnpc/capnproto"]
+	path = capnpc/capnproto
+	url = https://github.com/capnproto/capnproto.git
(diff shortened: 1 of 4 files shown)


zenhack:
I like this approach.

Bromeon (commented):
Randomly stumbled upon this PR and realized it has already been two years since the issue!
Very nice to see concrete ideas 🙂

PR #272, opened 2022-06-08
Include the name of the unimplemented method in the error message

nwalfield:
In the Server stubs that capnpc generates, don't just return "method
not implemented", say what method was not implemented to simplify
debugging.


nwalfield:
This results in more helpful errors messages like:

```
Error: Unimplemented: remote exception: method device::Server::list not implemented
```

dwrensha:
Thanks!

PR #273, opened 2022-06-23
The scope of the unsafe block can be appropriately reduced

peamaeq:


capnp/src/lib.rs
```diff
@@ -80,9 +80,9 @@ impl Word {
         let mut result: Vec<Word> = Vec::with_capacity(length);
         unsafe {
             result.set_len(length);
-            let p: *mut u8 = result.as_mut_ptr() as *mut u8;
-            core::ptr::write_bytes(p, 0u8, length * core::mem::size_of::<Word>());
         }
+            let p: *mut u8 = result.as_mut_ptr() as *mut u8;
```
dwrensha, line 84: the indentation is wrong here

capnp/src/private/layout.rs
```diff
@@ -2727,16 +2727,16 @@ impl <'a> PointerBuilder<'a> {
             wire_helpers::set_struct_pointer(
                 self.arena,
                 self.segment_id, self.cap_table, self.pointer, *value, canonicalize)?;
+            }
```
dwrensha, line 2730: indentation

capnp/src/private/layout.rs
```diff
@@ -2727,16 +2727,16 @@ impl <'a> PointerBuilder<'a> {
...
             Ok(())
-        }
     }
 
     pub fn set_list(&self, value: &ListReader, canonicalize: bool) -> Result<()> {
         unsafe {
             wire_helpers::set_list_pointer(self.arena, self.segment_id,
                                            self.cap_table, self.pointer, *value, canonicalize)?;
+                                        }
```
dwrensha, line 2738: indentation

capnp/src/private/layout.rs
```diff
@@ -3102,33 +3102,33 @@ impl <'a> StructBuilder<'a> {
             return Ok(())
         }
 
-        unsafe {
+
```
dwrensha, line 3105: If we remove this `unsafe` block then the following lines should be less indented.

dwrensha (changes requested):
Could you please fix up the formatting on these changes?

capnp/src/lib.rs
danieleades: see #305

Issue #221, opened 2021-03-13
no alloc support


theunkn0wn1:
I note #71 Which adds support for `no_std` crates, but `capnp` still depends on an allocator.
It is desirable to have support for environments that do not have an allocator / are memory constrained such as microcontrollers. 

In my application, I am presently using `postcard` to provide object de/encoding on both sides of the wire. Among other features, it supports de/encoding messages out of a `&[u8]` buffer, thus allowing it to work without alloc. 

I am evaluating capnproto to see if it can be a viable replacement for `postcard` and gRPC (replacing the latter with capnproto's RPC) as to simplify my software situation.

It appears that `capnp`, even in `no_std` mode, does not have a mechanism for statically parsing messages.

dwrensha:
Yes, that is accurate: the `capnp` crate still requires an allocator. I agree that it would be good to remove that requirement. To achieve that, we would need:
 1. An implementation of [`ReaderSegments`](https://github.com/capnproto/capnproto-rust/blob/eec64b6e55543fa4e6bdb047c49df957f764bdc3/capnp/src/message.rs#L93) that does not allocate
 2.  To remove the [`String`](https://github.com/capnproto/capnproto-rust/blob/eec64b6e55543fa4e6bdb047c49df957f764bdc3/capnp/src/lib.rs#L161) from `capnp::Error`.

Note, however, that getting the `capnp-rpc` crate (as opposed to just the base crate `capnp`) to work with `no_std` and no `alloc` would be a lot more difficult.

theunkn0wn1:
Thanks for the response.

For Capnproto to be viable, I don't require RPC support in no std; I only require the capability to statically decode a message of known maximum size.

romixlab:
Tried it on STM32F051 (thumbv6m) with alloc-cortex-m crate. Unfortunately something in `capnp::private::layout::wire_helpers::allocate ()` causes a HardFault to occur on `str	r1, [r5, #0]` instruction, while r1 = 0.

Btw size of the binary increased by approx 10kB after adding allocator and capnp dependencies (with opt-level = "z", codegen-units = 1, lto = true)

dwrensha:
We got another report of a hard fault using the alloc-cortex-m crate here: #293

dwrensha:
@romixlab I expect there's a good chance that if you upgrade to capnp-v0.14.10 the HardFault will go away (and probably be replaced by an allocation failure).

romixlab:
[...]

PR #276, opened 2022-07-07
capnp: adding capnp::serialize::NoAllocSliceSegments and capnp::serialize::read_message_from_flat_slice_no_alloc

tomkris:
I'm adding an implementation for parsing *unpacked* capnp message which does not require heap allocations on happy path (only for error message if error occurs).

# Context
no alloc support - #221 

Currently all existing options for parsing capnp message require heap allocations.

For example, even `SliceSegments` [will do heap allocation for storing segment offsets inside the `Vec`](https://github.com/capnproto/capnproto-rust/blob/master/capnp/src/serialize.rs#L43):

```
pub struct SliceSegments<'a> {
    words: &'a [u8],
    segment_indices : Vec<(usize, usize)>,
}
```

This is a problem for embedded applications where memory is scarce and heap allocations are either impossible or strongly discouraged, and having ability to parse message without doing heap allocations will improve experience of embedded developers using capnproto-rust.

In proposed implementation `alloc` crate is still required due to [use of `String` in `capnp::Error`](https://github.com/capnproto/capnproto-rust/blob/master/capnp/src/lib.rs#L161), however there will be no allocation happening on happy path, only in case of errors.

# Implementation details
The header of the [capnp message](https://capnproto.org/encoding.html#serialization-over-a-stream) has only list of segment sizes. Offsets of segments can be calculated by combining segment sizes which are coming before segment of interest. Because of this, noalloc implementation has to parse at least first N segment info in the header every time `NoAllocSliceSegments::get_segment` is called. This is a performance downside of using noalloc implementaiton.

To partially mitigate the performance issue, I implemented a special case for messages with a single segment - they go through a separate enum variant and won't require message parsing on `NoAllocSliceSegments::get_segment`.

Proposed implementation works only for messages which are serialized in "non-packed" format. Packed messages would need to be unpacked first (which requires heap allocation) and it's something that need to be addressed additionally.

## Considerations regarding reuse of `read_segment_table`

I attempted to reuse [read_segment_table](https://github.com/capnproto/capnproto-rust/blob/master/capnp/src/serialize.rs#L275) method in my implementation. Unfortunately it has [heap allocation inside it](https://github.com/capnproto/capnproto-rust/blob/master/capnp/src/serialize.rs#L311):
```
let mut segment_sizes = vec![0u8; (segment_count & !1) * 4];
```

Resulting `SegmentLengthsBuilder` [has `Vec`](https://github.com/capnproto/capnproto-rust/blob/master/capnp/src/serialize.rs#L193) inside it:

```
pub struct SegmentLengthsBuilder {
    segment_indices: Vec<(usize, usize)>,
    total_words: usize,
}
```

it also parses all segments, when in my case it's desired to only parse first N segments.

I tried to make logic work for alloc and noalloc cases through injected traits, but it become too cumbersome, I realized that benefit of reusing that method outweighs increased complexity of the logic, so I decided to implement all logic from scratch.

## Random number generation
I wanted to have a random number generator for generating test inputs. I introduced very naive random number generator, I was not sure if it's desired to have a build time dependency on [`rand` crate](https://crates.io/crates/rand), but  I'll be happy to make a change and take the dependency on it if it's your preference.

### DIFF
--- a/capnp/src/serialize.rs
+++ b/capnp/src/serialize.rs
@@ -23,6 +23,9 @@
 //! [standard stream framing](https://capnproto.org/encoding.html#serialization-over-a-stream),
 //! where each message is preceded by a segment table indicating the size of its segments.
 
+mod no_alloc_slice_segments;
+pub use no_alloc_slice_segments::NoAllocSliceSegments;
+
 use alloc::string::ToString;
 use alloc::vec::Vec;
 use core::convert::TryInto;
@@ -33,6 +36,7 @@ use crate::message;
 use crate::private::units::BYTES_PER_WORD;
 use crate::{Error, Result};
 
+pub const SEGMENTS_COUNT_LIMIT : usize = 512;
 
 /// Segments read from a single flat slice of words.
 pub struct SliceSegments<'a> {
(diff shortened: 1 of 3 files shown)


dwrensha:
Thanks! This is really cool. Did you do any performance measurement of the `SingleSegment` special case? In the past, when I've experimented with `smallvec`, I've found that sometimes the added branching overhead outweighs any benefits.

dwrensha:
> I wanted to have a random number generator for generating test inputs. I introduced very naive random number generator, I was not sure if it's desired to have a build time dependency on [rand crate](https://crates.io/crates/rand), but I'll be happy to make a change and take the dependency on it if it's your preference.

We already have a dependency on `quickcheck`, which depends on `rand`, so adding a new dev dependency on `rand` shouldn't actually pull in anything new. Can you make your tests just use quickcheck? I think that would be the simplest.

tomkris:
> We already have a dependency on `quickcheck`, which depends on `rand`, so adding a new dev dependency on `rand` shouldn't actually pull in anything new. Can you make your tests just use quickcheck? I think that would be the simplest.

Thanks, I'll make the change.

PR #279, opened 2022-07-19
capnp: use 'rand' in unit tests

tomkris:
Follow-up to https://github.com/capnproto/capnproto-rust/pull/276: use `rand` instead of housemade random number generator in tests.

I tried to figure out how to use `quickcheck` but don't have experience with it and did not figure out how exactly to use it.


dwrensha:
Thanks!

dwrensha:
I updated the tests to use quickcheck instead in https://github.com/capnproto/capnproto-rust/commit/7355a4ca590bbdb08435d177794746e3c9234207

PR #281, opened 2022-08-16
[experiment] try removing BufRead requirement in unpacking

dwrensha:
This is an experiment prompted by the tweet: https://twitter.com/nick_r_cameron/status/1559596640360693760.

Currently `serialize_packed::read_message()` has a `BufRead` bound. The question is: could we rewrite it to only use methods from the simpler `std::io::Read` trait and still achieve good performance?

At a first look, the answer seems to be "no". It is certainly true that the logic in `PackedRead::read()` can be simplified, but doing so has a noticeable effect on performance.

In particular, some of the benchmarks get ~25% slower. I suspect this is because in the simplified non-`BufRead` version the code does more branching.

```
$ cargo build --release -p benchmark
$ ./target/release/run_all_benchmarks ./target_release_benchmark
```

### DIFF
--- a/capnp/src/serialize_packed.rs
+++ b/capnp/src/serialize_packed.rs
@@ -23,22 +23,15 @@
 //! [packed stream encoding](https://capnproto.org/encoding.html#packing).
 
 use alloc::string::ToString;
-use core::{mem, ptr, slice};
-use crate::io::{Read, BufRead, Write};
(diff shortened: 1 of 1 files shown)


dwrensha:
Closing, because I don't see a way to make this work without a performance regression.

PR #284, opened 2022-08-28
Change list `get` methods to return an `Option<T>`

### LINKED ISSUE #251: Non-panicking `capnp::struct_list::Reader::get`
Hello!

Currently `capnp::struct_list::Reader::get` can panic if the index is out of bounds because of an assertion at https://docs.capnproto-rust.org/src/capnp/struct_list.rs.html#84

This is rather unexpected as the documentation doesn't say so and the usual convention for `get` method of collections is to return an `Option<T>` for in invalid index/key/etc.

Would it make sense to (1) change the prototype of `capnp::struct_list::Reader::get` to return an option and maybe (2) implement the `Index` trait with the old panicking behavior as done for slices in std?

I am willing to provide the PR, I am just unsure of what non backward compatible change are deemed acceptable.

marmeladema:
Hi @dwrensha 

I have opened this as a draft because its a work-in-progress and I have a few questions but I would like to start the discussion:

1. I have changed (some) `get` list reader methods to return an `Option<T>` and introduced a `get_unchecked` to mirror std slice API. This is a breaking change but as far as I know, the API is not considered stable at this point? I think that's the right call because its less confusing since it adhere to idiomatic rust practices but if the change is considered to severe, maybe adding a `checked_get` method instead would be more acceptable? A bonus point from this work is that in places where the index is already known to be in range, we can use `get_unchecked` directly and avoid some useless bound checks.

2. There are some `get` methods that already return a `Result<T>` but usually the index access is done *prior* to the call that returns the Result. In such case I believe the correct return type would be `Option<Result<T>>`. Do you agree?

3. What about `get` builder methods, should they be changed too for coherency?

Fixes #251

### DIFF
--- a/capnp/src/primitive_list.rs
+++ b/capnp/src/primitive_list.rs
@@ -62,13 +62,20 @@ impl <'a, T: PrimitiveElement> FromPointerReader<'a> for Reader<'a, T> {
 
 impl <'a, T: PrimitiveElement>  IndexMove<u32, T> for Reader<'a, T>{
     fn index_move(&self, index: u32) -> T {
-        self.get(index)
+        self.get(index).unwrap()
     }
 }
 
 impl <'a, T: PrimitiveElement> Reader<'a, T> {
-    pub fn get(&self, index: u32) -> T {
-        assert!(index < self.len());
+    pub fn get(&self, index: u32) -> Option<T> {
+        if index < self.len() {
+            Some(PrimitiveElement::get(&self.reader, index))
+        } else {
+            None
+        }
+    }
+
+    pub unsafe fn get_unchecked(&self, index: u32) -> T {
         PrimitiveElement::get(&self.reader, index)
     }
 
--- a/capnp/src/private/layout_test.rs
+++ b/capnp/src/private/layout_test.rs
@@ -120,16 +120,17 @@ fn bool_list() {
         let reader = crate::primitive_list::Reader::<bool>::get_from_pointer(&pointer_reader, None).unwrap();
 
         assert_eq!(reader.len(), 10);
-        assert_eq!(reader.get(0), true);
-        assert_eq!(reader.get(1), false);
-        assert_eq!(reader.get(2), true);
-        assert_eq!(reader.get(3), false);
-        assert_eq!(reader.get(4), true);
-        assert_eq!(reader.get(5), true);
-        assert_eq!(reader.get(6), true);
-        assert_eq!(reader.get(7), false);
-        assert_eq!(reader.get(8), false);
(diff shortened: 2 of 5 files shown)


dwrensha:
I suppose we could closely mirror the `std` API with:

* `get()` returns an `Option<T>`
* `unsafe get_unchecked()` returns. a `T`
* `index()` returns a `T` (and panics on out-of-bounds)

Existing `get()` calls would mostly need to migrate to `index()`. This makes things slightly more verbose, but it's not as bad as requiring `.get().unwrap()`.

I would want to roll this out as a multi-step process, where first we add `index()` and deprecate the current `get()`, and then on the next "major" version bump we change the type of `get()`.

This would cause some disruption to downstream users, but maybe it's worth it in the long run?

dwrensha:
> 2. There are some `get` methods that already return a `Result<T>` but usually the index access is done _prior_ to the call that returns the Result. In such case I believe the correct return type would be `Option<Result<T>>`. Do you agree?

Makes sense to me.

> 3. What about `get` builder methods, should they be changed too for coherency?

Yes, I think exactly the same concerns about out-of-bounds and panicking apply there too.

dwrensha:
Alternatively, we could keep the current `get()` methods as-is, and add new methods
- `fn try_get(self) -> Option<T>`
- `unsafe fn get_unchecked(self) -> T`

This would have the advantage of not disrupting any current users. Appending a `try_` to mean "can return None" is consistent with naming elsewhere in capnproto-rust, e.g. https://github.com/capnproto/capnproto-rust/blob/ec84c7ae76d3e655aff147f910f229c81e7028b6/capnp/src/serialize.rs#L266-L288

I'm not too worried about `get()` in capnp meaning something slightly different from `get()` in `std::collections`, especially if we document it.

marmeladema:
@dwrensha what do you prefer? Both approaches seem good to me. My main desire is to have *a* get-like function that does not panic. I personally think that introducing `try_get` is probably the easiest. No disruption for users and we can still decide later on if we want to change `get` itself.

marmeladema:
Opened https://github.com/capnproto/capnproto-rust/pull/285 for the `try_get` approach

dwrensha:
Yeah `try_get` sounds good to me. We can rename the methods later if that seems like the appropriate thing.

dwrensha:
Added some documentation: https://github.com/capnproto/capnproto-rust/commit/da7a9f7163ef64c888f1a425e18e9e991ceba1c5

dwrensha:
Closing in favor of #285. Please open a new pull request if you want to add `get_unchecked()` methods.

PR #285, opened 2022-08-29
Introduce `try_get` list methods

### LINKED ISSUE #251: Non-panicking `capnp::struct_list::Reader::get`

marmeladema:
@dwrensha here is the version with `try_get` methods

Fixes #251

### DIFF
--- a/capnp/src/text_list.rs
+++ b/capnp/src/text_list.rs
@@ -68,6 +68,14 @@ impl <'a> Reader<'a> {
         assert!(index <  self.len());
         self.reader.get_pointer_element(index).get_text(None)
     }
+
+    pub fn try_get(self, index : u32) -> Option<Result<crate::text::Reader<'a>>> {
+        if index <  self.len() {
+            Some(self.reader.get_pointer_element(index).get_text(None))
+        } else {
+            None
+        }
+    }
 }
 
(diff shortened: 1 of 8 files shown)


capnp/src/text_list.rs
```diff
@@ -117,8 +125,17 @@ impl <'a> FromPointerBuilder<'a> for Builder<'a> {
 
 impl <'a> Builder<'a> {
     pub fn get(self, index: u32) -> Result<crate::text::Builder<'a>> {
+        assert!(index <  self.len());
```
dwrensha, line 128: good catch!

dwrensha:
Thanks!

PR #286, opened 2022-08-30
Avoid manual indexing of slices in for loops

marmeladema:
This is replaced by either iterating over the elements of the slice
or by calling the appropriate slice method.

### DIFF
--- a/capnp/src/serialize.rs
+++ b/capnp/src/serialize.rs
@@ -391,9 +391,7 @@ fn flatten_segments<R: message::ReaderSegments + ?Sized>(segments: &R) -> Vec<u8
     }
     for i in 0..segment_count {
         let segment = segments.get_segment(i as u32).unwrap();
-        for idx in 0..segment.len() {
-            result.push(segment[idx]);
-        }
+        result.extend(segment);
     }
     result
 }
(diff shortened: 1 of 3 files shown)


marmeladema:
`slice::fill` was stabilized in rust 1.50, a year and a half ago but the minrust job uses 1.44.
@dwrensha whats the policy for increasing MSRV? It's not clearly documented in the README but what do you think about updating to 1.56 to be able to migrate to edition 2021 at some point?

dwrensha:
My policy has been to only update the minimum required Rust version when we make "major" releases -- the next one would be when we go from 0.14 to 0.15. I'm not sure when that will be, but the more reasons like this that accumulate, the more motivated I am to make it happen.

Is there anything in particular from edition 2021 that you want?

marmeladema:
Alright, I removed the calls to the `fill` method. Should be ready for review now.

About edition 2021, there are things that I find quite useful:
* https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html
* https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html

 but I don't have a good idea if they are interesting for capnproto-rust itself. I proposed rust 1.56 just because it was recent enough but not too recent either and brings some potentially useful improvements.

dwrensha:
Thanks!

PR #287, opened 2022-08-31
Remove needless lifetime annotation

marmeladema:
Another round of small clippy improvements.


marmeladema:
@dwrensha I am trying to fix most clippy lints that I find useful. I hope you don't mind.
This one is imo making the codebase more readable.

PR #275, opened 2022-06-28
Bump quickcheck to 1

dkg:
This attempts to address #274 

Feedback welcome!  I'm not sure whether this warrants a semver-incompatible version bump.

PR #290, opened 2022-08-31
Add `TYPE_ID` associated constant to `HasTypeId` trait

marmeladema:


marmeladema:
I have some need to be able to manipulate the type id as a constant therefor I added a `TYPE_ID` associated constant to the `HasTypeId` trait as functions in trait cannot be const at the moment.

dwrensha:
Does this actually let you use the value where you want to? In https://github.com/capnproto/capnproto-rust/pull/225 I found that Rust was more restrictive than I expected about where associated constants can be used.

marmeladema:
@dwrensha  yes, it works perfectly fine in my case.

dwrensha:
I like this change, but it will require a version bump. I'm thinking I'll do the version bump next week, pulling in this, #225, and #275.

marmeladema:
Thank you! That works for me.
Just for my sake, why would it require a (major?) version bump? Because the trait is public and potentially implemented by downstream types so adding an associated constant would break them?

dwrensha:
Ah, I was assuming that you were removing the `type_id()` method, but I see now that you're keeping it and adding a default impl. Yeah, that's less risky, and we could probably get away with not bumping the major version.

dwrensha:
Hm... I agree that this probably won't cause a problem for anyone who generates code just-in-time in a `build.rs`. However, some users generate code once beforehand and check in the results. (We in fact follow that pattern ourselves for the capnp-rpc crate.) This change will force such users to update their `capnpc` crate version and to regenerate their code. That feels to me like something that should only happen on a "major" release of `capnp`.

marmeladema:
I see, that makes perfect sense.

Alternatively, if we aren't ready to do a major version bump, we could instead declare the constant in each type module (the one that contains `Owned` / `Reader` and `Builder` structs), remove it from the nested `_private` module and change the `HasTypeId` generated code to use that moved constant instead of `_private`.
This way, it's fully backward compatible, but still exposes a const variable. It would not be tied to a trait but it would still be useful in my case.

We could always add the trait associated constant at a later point when we'ready for the major bump.

dwrensha:
Looks like Generic Associated Types are likely to be arriving soon: https://github.com/rust-lang/rust/pull/96709
I'm inclined to say that we should delay our major version bump until then, so that we can roll in the cleanup of `Owned` and `OwnedStruct` that will be unblocked.

dwrensha:
Merging now, to be included in the 0.15.0 release, which I intend to make after Rust 1.65 comes out next week.

PR #292, opened 2022-09-02
Make `IndexMove` private and unsafely implement it for all lists

marmeladema:
In some sense it acts as a `get_unchecked` method but is not
exposed publicly anymore to prevent rogue implementations.


dwrensha:
If I understand correctly, the purpose of this is to make it so that iterator-based access to capnproto lists avoids bounds checks. Have you experimentally observed a significant improvement from this change? I'm somewhat reluctant to add more instances of `unsafe` like this.

PR #294, opened 2022-09-22
Change capnp::io::Write trait to return the number of written bytes on success

guillaume-michel:
This PR change the `capnp::io::Write` trait to return the number of written bytes on success. It is particularly useful when using `serialize_packed::write_message` on embedded platform where we usually serialize to an array then send the serialized bytes over the wire (like UART, SPI, ...).

We usually end up with the following code:
```rust
let mut buf: [u8; 32] = [0; 32];

capnp::serialize_packed::write_message(&mut buf.as_mut_slice(), &message).unwrap();

// with current capnp, the only way to retrieve the size of the packed message
// is copy and modify PackedWrite struct to record the amount of written bytes
// and copy and modify write_message function to return the written bytes.      
```
with the following PR, the amount of written bytes is returned by `write_message` function and used like this:
```rust
let mut buf: [u8; 32] = [0; 32];

let written_bytes = capnp::serialize_packed::write_message(&mut buf.as_mut_slice(), &message).unwrap();

send_to_uart(buf[..written_bytes]);

```

### DIFF
--- a/capnp/src/io.rs
+++ b/capnp/src/io.rs
@@ -42,7 +42,7 @@ pub trait BufRead : Read {
 
 /// A rough approximation of std::io::Write.
 pub trait Write {
-    fn write_all(&mut self, buf: &[u8]) -> Result<()>;
+    fn write_all(&mut self, buf: &[u8]) -> Result<usize>;
 }
 
 #[cfg(feature="std")]
@@ -72,9 +72,9 @@ mod std_impls {
     }
 
     impl <W> Write for W where W: std::io::Write {
-        fn write_all(&mut self, buf: &[u8]) -> Result<()> {
+        fn write_all(&mut self, buf: &[u8]) -> Result<usize> {
             std::io::Write::write_all(self, buf)?;
-            Ok(())
+            Ok(buf.len())
         }
     }
 }
@@ -86,7 +86,7 @@ mod no_std_impls {
     use crate::io::{Read, BufRead, Write};
 
     impl <'a> Write for &'a mut [u8] {
-        fn write_all(&mut self, buf: &[u8]) -> Result<()> {
+        fn write_all(&mut self, buf: &[u8]) -> Result<usize> {
             if buf.len() > self.len() {
                 return Err(Error::failed("buffer is not large enough".to_string()));
             }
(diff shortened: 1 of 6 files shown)


example/addressbook/addressbook.rs
```diff
@@ -63,7 +63,11 @@ pub mod addressbook {
             }
         }
 
-        serialize_packed::write_message(&mut ::std::io::stdout(), &message)
+        let written_bytes = serialize_packed::write_message(&mut ::std::io::stdout(), &message)?;
+
+        assert_eq!(written_bytes, 151);
```
guillaume-michel, line 68: This is the size of the message when dump to disk whith the following command:
```bash
./target/debug/addressbook write > /tmp/message.bin
ls -l /tmp/message.bin | awk '{print $5}'
```

capnp/src/io.rs
```diff
@@ -42,7 +42,7 @@ pub trait BufRead : Read {
 
 /// A rough approximation of std::io::Write.
 pub trait Write {
-    fn write_all(&mut self, buf: &[u8]) -> Result<()>;
+    fn write_all(&mut self, buf: &[u8]) -> Result<usize>;
```
dwrensha, line 45: The `usize` return value does not give us any new information. An `Ok` result will always have written exactly `buf.len()` bytes.

guillaume-michel: If I am not mistaken, it is not the case when using packed serialization.

guillaume-michel:
I am not an expert in Rust but adding capnp::io::Cursor similar to std::io::Cursor, we could use the position() method to retrieve the number bytes written in a much more elegant way. It will also avoid having breaking changes in the API.

Unfortunately, I am not sure I am able to implement it for now.

dwrensha:
Note also that the `capnp::io::Write` implementation for `&mut [u8]` updates the slice's starting position: https://github.com/capnproto/capnproto-rust/blob/32ea1c34d3f88aa73894271d9b5a7c3292fc1dee/capnp/src/io.rs#L94-L96

So you can recover the number of bytes written like this:

```rust
let mut buf: [u8; 32] = [0; 32];
let mut buf_slice = buf.as_mut_slice();
capnp::serialize_packed::write_message(&mut buf_slice, &message).unwrap();
let bytes_written = buf.len() - buf_slice.len();
```

guillaume-michel:
Even better! I guess I still have more to learn. You can close this PR as it is not needed anymore. Thanks for you help

PR #305, opened 2022-11-11
use auto-formatting

danieleades:
the benefits of auto-formatting are many and varied, and are well-documented

summarily,

- it creates a consistent visual style across the ecosystem (rustfmt is almost universally adopted)
- reduces diff noise (primarily by allowing a contributor to autoformat their changes without affecting other code)
- **obviates the need to ever discuss formatting in a PR again**

I would personally strongly advocate for using the default config, but if there are any consistent differences in style from the default then these can be configured.


dwrensha:
Is there a way to excluded the generate files `capnpc/src/schema_capnp.rs`, `capnp-rpc/src/rpc_capnp.rs` and `capnp-rpc/src/rpc_twoparty_capnp.rs` from this formatting?

danieleades:
> Is there a way to exclude the generate files `capnpc/src/schema_capnp.rs`, `capnp-rpc/src/rpc_capnp.rs` and `capnp-rpc/src/rpc_twoparty_capnp.rs` from this formatting?

i'll look into it. Why these files specifically?

dwrensha:
Nice, and thanks @Timmmm for adding `@generated` in #153!

PR #313, opened 2022-11-11
remove explicit lifetimes where they can be elided

danieleades:


dwrensha:
I prefer the explicit lifetime style, especially since we sometimes cast from unsafe pointers. In the past I've been bitten by lifetime elision doing unexpected things (particularly where then are trait objects involved). I think we should suppress this Clippy lint.

danieleades:
> I prefer the explicit lifetime style, especially since we sometimes cast from unsafe pointers. In the past I've been bitten by lifetime elision doing unexpected things (particularly where then are trait objects involved). I think we should suppress this Clippy lint.

can you give an example where this has bitten you? I'm struggling to understand how this can cause confusion. Can you show me some code?

In the cases in this PR the lifetimes are not needed because they are unambiguous. As soon as it becomes ambiguous (due to the addition of a second lifetime, or something) then there's a compiler error and you're forced to add explicit lifetimes.

dwrensha:
I don't really remember my examples. They were from like 8 years ago. I suppose Rust may have fixed some bugs since then, and I should stop being a curmudgeon about this.

PR #316, opened 2022-11-11
add dependabot config

danieleades:


dwrensha:
What does this do? Does this mean that we will get automatic issues like https://github.com/sandstorm-io/collections-app/pull/20 ? My initial impression is that sounds more annoying than useful.

danieleades:
yes, dependabot creates automatic PRs for version bumps of dependencies. It also handles automatic rebasing and removal of the PRs in response to changes due to other PRs, as appropriate.

dependabot has the potential to be noisy for *binaries* since even a patch change in the dependency requires an update to the cargo.lock file. In the case of libraries, dependabot will only trigger a PR in the case of a major change. This repo is a library, so I wouldn't expect dependabot to be noisy here. It's also possible to configure dependabot to only run scans at fixed intervals, say monthly. Don't think that's necessary here for the above reasons, but that's available to you if you do find it annoying.

If dependabot is opening PRs, it's because you've got out of date dependencies that need updating. Personally, I find it extremely useful.

danieleades:
@dwrensha I had a look into this, and as it stands I think dependabot would open  precisely 3 PRs to update the `tokio_util` dependency in each of the `capnp-rpc` examples. These would then pass CI and could be merged without further effort.

There are no other outdated dependencies, so that would be the end of it, until one of the deps had a major update.

dwrensha:
Ok, I'm willing to try this.

PR #317, opened 2022-11-11
remove needless pass by value

danieleades:
i'll leave this as a draft for now. I'm on shakier ground with this one, and this is probably (i've not actually checked) a breaking change


dwrensha:
Ah, looks like this now longer introduces any breaking changes. Thanks!

PR #318, opened 2022-11-11
remove manual implementation of 'AddAssign'

danieleades:


dwrensha:
This was a breaking change. See #332, which re-adds the old method and deprecates it. When we release 0.16.0 we can remove it.

PR #319, opened 2022-11-11
remove redundant 'extern crate' calls

danieleades:
I'll leave this as a draft. I don't *believe* these are required, but i'm not 100% sure there isn't some `no-std`/`alloc` weirdness

### DIFF
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -76,7 +76,7 @@ impl CodeGenerationCommand {
     }
 
     /// Generates Rust code according to a `schema_capnp::code_generator_request` read from `inp`.
-    pub fn run<T>(&mut self, inp: T) -> ::capnp::Result<()>
+    pub fn run<T>(&mut self, inp: T) -> capnp::Result<()>
         where T: std::io::Read
     {
         use capnp::serialize;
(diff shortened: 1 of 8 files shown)


danieleades:
noticing some quirks in the code-gen. All imports are made using a `::` prefix, which means that the deps must be explicitly imported in the crate root, which generally isn't done since 2015 edition.

Also, the code-gen is *very* bespoke. Any reason for not using something like [quote](https://docs.rs/quote/latest/quote/)?

dwrensha:
The module for the generated code can be configured via an attribute: https://github.com/capnproto/capnproto-rust/blob/4083728a4b1f1a4fb750cf07d03c078e8fec3c45/capnpc/rust.capnp#L17-L28

danieleades:
> The fully-qualified module paths are to minimize the chance of name collisions. See e.g. #77

i understand why fully qualified paths have been used, but i don't understand why they need to lead with `::capn...` instead of `capn...`. The former requires that `capn` is explicitly imported in the crate root, which is not idiomatic. Is it strictly necessary?

danieleades:
> My understanding is that in the 2015 edition, the `extern crate` is always needed, regardless of whether we have `capnp::...` or `::capnp::...`.

agreed

> For later editions, I don't think there are any problems with our current approach.

forcing consumers to import capnp in their crate root, even if they're not directly using it is a teensy bit unexpected.
For me that weirdness is worse than the risk of someone trying to import two crates both called `capnp` (which they would have to explicitly disambiguate *anyway*), but I guess it's a matter of taste how you weight those two things.

I guess i'd better get this PR to actually build first, and then we can discuss the merits? If i can't get it to work then the point is moot...

dwrensha:
> forcing consumers to import capnp in their crate root

I'm not sure what exactly you mean by this. As far as I know, consumers in the 2018 and 2021 editions do not need to add an `extern crate` or a `use capnp` statement or anything like that in the crate root.

Do you mean that they need to add the dependency to `capnp` in their `Cargo.toml`? I don't see how we could avoid that.

danieleades:
> Do you mean that they need to add the dependency to `capnp` in their `Cargo.toml`? I don't see how we could avoid that.

no i agree that's not avoidable

i'll close this PR, i think i misunderstood that some of the tests were specifically checking the 2015-edition case

PR #325, opened 2022-11-12
refactor 'Word' struct

danieleades:
this PR refactors the `capnp::word` method to be an associated method of the `capnp::Word` struct.

Additionally, the constructor is updated to accept an array directly. This allows for more compact formatting, and allows using array initialiser shorthand.


dwrensha:
> this PR refactors the capnp::word method to be an associated method of the capnp::Word struct.

`::capnp::word()` can show up a lot in generated code (for consts and default values), so I'm concerned that making it more verbose like this will bloat the generated (source) code.

danieleades:
> > this PR refactors the capnp::word method to be an associated method of the capnp::Word struct.
> 
> `::capnp::word()` can show up a lot in generated code (for consts and default values), so I'm concerned that making it more verbose like this will bloat the generated (source) code.

Is it more verbose?

It's changed from `::capnp::word` to `::capnp::Word::new`. That seems pretty cheap, no?

dwrensha:
Maybe we could keep `capnp::word()` but also add `Word::new([u8;8])`? That way we would also avoid making any breaking changes.

danieleades:
> Maybe we could keep `capnp::word()` but also add `Word::new([u8;8])`? That way we would also avoid making any breaking changes.

sure thing

danieleades:
> Is `clippy::too_many_arguments` the only lint that's getting triggered by `capnp::word()`? Or does Clippy also have a lint that wants to move it from a standalone function into `Word::new()` method?

it's the only lint that's affected.

Moving the standalone function to an associated method is an aesthetic choice from me. It's essentially a constructor, so i don't see really see why it should be treated differently to all other 'constructors' in this library. I seem to recall there was a brief period when only free functions could be `const`? Maybe that has something to do with it.

dwrensha:
> It's changed from ::capnp::word to ::capnp::Word::new. That seems pretty cheap, no?

Someday I hope we'll be storing the entire encoded schema in the generated Rust files, similar to how c++ does it, e.g.:

https://github.com/capnproto/capnproto/blob/4d4832c2bdd3c9c218bbaef462e7767c314bf426/c%2B%2B/src/capnp/schema.capnp.c%2B%2B#L9-L233

This makes it possible to support a ["dynamic types"](https://capnproto.org/cxx.html#dynamic-reflection) API which can be useful for lots of things.

5 extra bytes per word, multiplied by thousands of words in a single schema, is not nothing.

danieleades:
If the number of characters is a concern there are probably better approaches than using fully-qualified paths for every object. A better approach might be to generate the implementations inside a new scope, and export any objects you need to be public.

instead of this-

```rust
#[derive(Clone, Copy)]
pub struct Reader<'a> {
    reader: ::capnp::private::layout::StructReader<'a>,
}

impl<'a> ::capnp::traits::HasTypeId for Reader<'a> {
    const TYPE_ID: u64 = _private::TYPE_ID;
}
impl<'a> ::capnp::traits::FromStructReader<'a> for Reader<'a> {
    fn new(reader: ::capnp::private::layout::StructReader<'a>) -> Reader<'a> {
        Reader { reader }
    }
}

impl<'a> ::capnp::traits::FromPointerReader<'a> for Reader<'a> {
    fn get_from_pointer(
        reader: &::capnp::private::layout::PointerReader<'a>,
        default: ::core::option::Option<&'a [capnp::Word]>,
    ) -> ::capnp::Result<Reader<'a>> {
        ::core::result::Result::Ok(::capnp::traits::FromStructReader::new(
            reader.get_struct(default)?,
        ))
    }
}

impl<'a> ::capnp::traits::IntoInternalStructReader<'a> for Reader<'a> {
    fn into_internal_struct_reader(self) -> ::capnp::private::layout::StructReader<'a> {
        self.reader
    }
}
```

you could do this-

```rust
mod _private_module {
    use capnp::{
        private::layout::{PointerReader, StructReader},
        traits::{FromPointerReader, FromStructReader, HasTypeId, IntoInternalStructReader},
        Result, Word,
    };
    use core::option::Option;

    #[derive(Clone, Copy)]
    pub struct Reader<'a> {
        reader: StructReader<'a>,
    }

    impl<'a> HasTypeId for Reader<'a> {
        const TYPE_ID: u64 = _private::TYPE_ID;
    }
    impl<'a> FromStructReader<'a> for Reader<'a> {
        fn new(reader: StructReader<'a>) -> Reader<'a> {
            Reader { reader }
        }
    }

    impl<'a> FromPointerReader<'a> for Reader<'a> {
        fn get_from_pointer(
            reader: &PointerReader<'a>,
            default: Option<&'a [Word]>,
        ) -> Result<Reader<'a>> {
            Result::Ok(FromStructReader::new(reader.get_struct(default)?))
        }
    }

    impl<'a> IntoInternalStructReader<'a> for Reader<'a> {
        fn into_internal_struct_reader(self) -> StructReader<'a> {
            self.reader
        }
    }
}

pub use _private_module::Reader;
```

dwrensha:
Thanks for the suggestion, but I think the existing `capnp::word()` function works well enough. `capnp::Word::new()` makes things more verbose, and introduces needless API churn.

PR #333, opened 2022-11-15
implement 'is_empty' where it's useful,

danieleades:
I've globally suppressed the [clippy::len_without_is_empty](https://rust-lang.github.io/rust-clippy/master/#len_without_is_empty) lint, and implemented it on the arena type which actually makes use of a length comparison to 0.

I would personally prefer to suppress the lint separately on each `Builder` and `Reader` struct, rather than a blanket suppression, but I wanted to get your feedback


danieleades:
changed my mind. Have moved the lint suppression to the methods. It's a noisier diff now, but i think it's worth it to minimise the scope of the suppression

dwrensha:
Instead of suppressing the lint, can we add the missing `is_empty()` methods instead? I don't see a reason not to.

PR #337, opened 2022-11-19
remove 'FromStructReader' trait

danieleades:
`FromStructReader` is equivalent to `From<StructReader>`. The latter is preferred, since it lets you use the standard Rust `From` and `Into` machinery


dwrensha:
Done in https://github.com/capnproto/capnproto-rust/commit/0bfb8d711110d8c3717215f34d9ddd8ee23ab6e3

PR #340, opened 2022-11-19
use built-in 'Default' trait

danieleades:
- use the built-in `Default` trait for constructing structs with no arguments
- use idiomatic `new` method for constructing structs with arguments

### DIFF
--- a/capnp/src/message.rs
+++ b/capnp/src/message.rs
@@ -51,7 +51,7 @@
 //! use capnp::message::{self, TypedBuilder, TypedReader};
 //!
 //! fn main() {
-//!     let mut builder = TypedBuilder::<simple_struct::Owned>::new_default();
+//!     let mut builder = TypedBuilder::<simple_struct::Owned>::default();
 //!     let mut builder_root = builder.init_root();
 //!     builder_root.set_x(10);
 //!     builder_root.set_y(20);
@@ -462,7 +462,7 @@ where
 
     pub fn get_root_as_reader<'a, T: FromPointerReader<'a>>(&'a self) -> Result<T> {
         if self.arena.is_empty() {
-            any_pointer::Reader::new(layout::PointerReader::new_default()).get_as()
+            any_pointer::Reader::new(layout::PointerReader::default()).get_as()
         } else {
             let (segment_start, _segment_len) = self.arena.get_segment(0)?;
             let pointer_reader = layout::PointerReader::get_root(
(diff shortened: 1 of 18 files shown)


capnp/src/message.rs
```diff
@@ -708,10 +708,10 @@ fn test_allocate_max() {
     allocator.deallocate_segment(a3, s3, 0);
 }
 
-impl Builder<HeapAllocator> {
+impl Default for Builder<HeapAllocator> {
     /// Constructs a new `message::Builder<HeapAllocator>` whose first segment has length
     /// `SUGGESTED_FIRST_SEGMENT_WORDS`.
-    pub fn new_default() -> Builder<HeapAllocator> {
```
dwrensha, line 714: Abruptly deleting this method would cause unnecessary pain. It would be better to keep it around and deprecate it with a warning like "use the default() method instead (from the Default trait)".

capnp/src/message.rs
```diff
@@ -526,12 +526,12 @@ where
     message: Builder<A>,
 }
 
-impl<T> TypedBuilder<T, HeapAllocator>
+impl<T> Default for TypedBuilder<T, HeapAllocator>
 where
     T: Owned,
 {
-    pub fn new_default() -> Self {
```
dwrensha, line 533: Abruptly deleting this method would cause unnecessary pain. It would be better to keep it around and deprecate it with a warning like "use the default() method instead (from the Default trait)".

capnp/src/message.rs
danieleades: in what version should each of these be deprecated?

danieleades: ping @dwrensha

danieleades:
i've updated this PR in light of the fact that the other breaking changes have been merged.

capnp/src/message.rs
laundmo: i think they meant the deprecated hint: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

danieleades: > i think they meant the deprecated hint: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

exactly. This attribute has the `since` argument, which defines the version in which it should be deprecated. Hence my question.

laundmo: wouldn't that just be the current version upon release?

danieleades:
@dwrensha i've refreshed this PR, and deprecated the existing methods rather than removing them outright

danieleades:
@dwrensha is this still of interest?

dwrensha:
I added `impl Default for message::Builder<HeapAllocator>` in https://github.com/capnproto/capnproto-rust/commit/2018225c26612fa77e95405c237cacdc671c3719 because that seems potentially helpful if users are defining their own structs that include a `message::Builder` and they want to have `impl Default` on those structs. I don't want to remove `message::Builder::new_default()` because requiring users to find a trait method seems like a potential source of confusion. The other changes here are to internal details and don't seem like an improvement to me.

PR #341, opened 2022-11-19
Ported reconnect API

griff:
This is more or less a straight port of the reconnect API from capnp-c++.

I say more or less because I also needed the `setTarget` functionality from `CapRedirector` in Sandstorm and so have included that as part of the API.

To keep the API consistent with the existing `new_client` and `new_promise_client` functions the `auto_reconnect` and `lazy_auto_reconnect` functions work with generic clients that implement `FromClientHook`.


dwrensha:
Thanks!

PR #342, opened 2022-11-20
mark unsafe methods

danieleades:
see https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref

### DIFF
--- a/capnp/src/message.rs
+++ b/capnp/src/message.rs
@@ -377,7 +377,7 @@ pub unsafe trait Allocator {
     /// `word_size` is the length of the segment in words, as returned from `allocate_segment()`.
     /// `words_used` is always less than or equal to `word_size`, and indicates how many
     /// words (contiguous from the start of the segment) were possibly written with non-zero values.
-    fn deallocate_segment(&mut self, ptr: *mut u8, word_size: u32, words_used: u32);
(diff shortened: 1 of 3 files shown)


danieleades:
looking at the signature of `deallocate_segment` it's kind of wild that this method is *not* already marked as unsafe. I imagine this change provides some impetus to cutting a new release in the near future?

dwrensha:
That's a method that's not intended to be called by end users. We have a lot of other such methods under the `private` module that arguably should also be marked `unsafe`, e.g. `PointerReader::get_root()`, but I'm not too worried about any of them because it's pretty clear that they are not intended as public API.

For now, I think expanding the doc comment on `deallocate_segment()` would be sufficient.

dwrensha:
(I do think we should clean these things up, but I don't think it's urgent enough that we need to rush out a new release.)

danieleades:
> because it's pretty clear that they are not intended as public API

surely this means that correctly marking them as unsafe is then *not* a breaking change?

dwrensha:
> surely this means that correctly marking them as unsafe is then not a breaking change?

Users are supposed to be able to implement the `Allocator` trait, but are not supposed to call it directly. This change is breaking for uses implement the `Allocator` trait.

dwrensha:
rebased and merged in https://github.com/capnproto/capnproto-rust/commit/1d4c541604a2788dfaee41ae5f31322781e5b85d

PR #343, opened 2022-11-21
Clippy/general

danieleades:
more general clippy fixes.

This PR is based on #342

I've discovered that Clippy was erroring out early due to #342, so with these errors addressed a whole bunch more warnings are visible.


danieleades:
@dwrensha this is ready for review. Might be easiest to review the commits one at a time.

Note that this PR also includes the commit in #342

danieleades:
@dwrensha this has been rebased. Any changes required before merging?

PR #345, opened 2022-11-21
remove unnecessary use of 'ref'

danieleades:
'ref' is a little overused. This PR strips out some unnecessary uses


danieleades:
> The reason for the `ref`s is that most of this code was written before [match ergonomics](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html) existed

Ah that makes sense

PR #349, opened 2022-11-21
remove 'IntoListReader' trait

danieleades:


dwrensha:
This trait appears in user-facing API here: https://github.com/capnproto/capnproto-rust/blob/636e4d50af63efd2f61dc8e08c7573c71420af75/capnp/src/raw.rs#L50

After this change, the private `ListReader` type would appear there instead, which I would prefer not to expose in such an API. I don't want users clicking on documentation links trying to understand this function, and ending up in the `capnp::private` namespace, where they usually don't need to be.

I've added a bit of documentation here: https://github.com/capnproto/capnproto-rust/commit/636e4d50af63efd2f61dc8e08c7573c71420af75

danieleades:
> This trait appears in user-facing API here:
> 
> https://github.com/capnproto/capnproto-rust/blob/636e4d50af63efd2f61dc8e08c7573c71420af75/capnp/src/raw.rs#L50
> 
> After this change, the private `ListReader` type would appear there instead, which I would prefer not to expose in such an API. I don't want users clicking on documentation links trying to understand this function, and ending up in the `capnp::private` namespace, where they usually don't need to be.
> 
> I've added a bit of documentation here: [636e4d5](https://github.com/capnproto/capnproto-rust/commit/636e4d50af63efd2f61dc8e08c7573c71420af75)

Can you not hide the type using `#[doc(hidden)]`?

PR #351, opened 2022-11-22
remove 'IntoInternalStructReader' trait

danieleades:


dwrensha:
My comment about `IntoInternalListReader` also applies here. https://github.com/capnproto/capnproto-rust/pull/349#issuecomment-1384536651

PR #362, opened 2023-01-11
Refactor builder arena

marmeladema:
Each commit should be self explainable.
This makes the code simpler and should help the compiler optimize things a bit better.

### DIFF
--- a/capnp/src/private/arena.rs
+++ b/capnp/src/private/arena.rs
@@ -175,6 +175,18 @@ struct BuilderSegment {
     allocated: u32, // in words
 }
 
+impl BuilderSegment {
+    fn allocate(&mut self, amount: WordCount32) -> Option<u32> {
+        if amount > self.capacity - self.allocated {
+            None
+        } else {
+            let result = self.allocated;
+            self.allocated += amount;
+            Some(result)
+        }
+    }
+}
+
 pub struct BuilderArenaImplInner<A>
 where
     A: Allocator,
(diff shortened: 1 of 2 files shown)


marmeladema:
There is a small bug in the first commit because the vec of segments is leaked. It is not strictly unsound bit I'll fix it tomorrow anyway.

marmeladema:
@dwrensha this is ready for review :)

Unless `capnp/src/private` is officially part of the public API, it should not be a breaking change.

capnp/src/private/arena.rs
```diff
@@ -288,54 +307,37 @@ where
...
-            None => unreachable!(),
-        };
+    fn allocate_segment(&mut self, minimum_size: WordCount32) -> SegmentId {
+        let seg = self.allocator.allocate_segment(minimum_size);
+        let segment_id = self.segments.len() as u32;
         self.segments.push(BuilderSegment {
             ptr: seg.0,
             capacity: seg.1,
-            allocated: 0,
+            allocated: minimum_size,
```
dwrensha, line 316: This looks wrong to me. `allocated` is supposed to indicate how many words of the segment are already in use, which should be zero when the segment is first allocated. It seems to me that what you have here would result in a lot of unused memory.

I pushed some new doc comments for `BuilderSegment` that might make things clearer: https://github.com/capnproto/capnproto-rust/commit/888844337f000aeebd7b8fbb7d1bfcc6e728acd5

I suppose one things that's potentially confusing about all this is that the word "allocate" is being used for different things:
  1. the `message::Allocator`, whose job is to provide memory from the system
  2. `BuilderArena::allocate()`, which takes takes memory already-"allocated" memory from (1) and reserves it for usage by `private::layout`.

dwrensha: Oh, I see. This has to do with commit https://github.com/capnproto/capnproto-rust/pull/362/commits/b6f69313d375fcdeb16c44a015b7f14deaff4881. It happens to work because there are only a two usages of the `allocate_segment()` method, and they follow a particular pattern.

I'd prefer not to make this change. If we do make it, we should at least add some comments about the new behavior, and maybe also change some naming.

marmeladema: Maybe we can introduce a dedicated method for this then? I think having a non fallible method helps with readability and optimizations

capnp/src/private/arena.rs
```diff
@@ -247,7 +261,12 @@ where
     pub fn into_allocator(self) -> A {
         let mut inner = self.inner.into_inner();
         inner.deallocate_all();
-        inner.allocator.take().unwrap()
+        // See https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/.E2.9C.94.20Extracting.20field.20from.20struct.20that.20implement.20.60Drop.60
+        let (allocator, _segments) = unsafe {
```
dwrensha, line 265: I'm hesitant to add more `unsafe` like this. I'm tempted to pull in the changes from ##363 *without* the changes from this PR, to see if that still works well. Getting rid of that `RefCell` to me is the big win of all this.

marmeladema: As you wish :-) I will likely be testing all those changes in production in the next couple of weeks.

I do think simplifying the codebase, is worth one unsafe block though.

dwrensha: https://github.com/capnproto/capnproto-rust/pull/365

PR #363, opened 2023-01-11
[RFC] Remove `RefCell` from builder arena

marmeladema:
Based on https://github.com/capnproto/capnproto-rust/pull/362 so only the last commit is important here.

The usage of `RefCell` was noticeable in production flamegraphs I gathered and since it is not strictly required, this proposal attempts to remove it.

There are probably other ways but this way was pretty straightforward, even though, very invasive.

### DIFF
--- a/capnp/src/private/arena.rs
+++ b/capnp/src/private/arena.rs
@@ -175,6 +175,18 @@ struct BuilderSegment {
     allocated: u32, // in words
 }
 
+impl BuilderSegment {
+    fn allocate(&mut self, amount: WordCount32) -> Option<u32> {
+        if amount > self.capacity - self.allocated {
+            None
+        } else {
+            let result = self.allocated;
+            self.allocated += amount;
+            Some(result)
+        }
+    }
+}
+
 pub struct BuilderArenaImplInner<A>
 where
     A: Allocator,
@@ -305,31 +317,24 @@ where
     }
 
     fn allocate(&mut self, segment_id: u32, amount: WordCount32) -> Option<u32> {
-        let seg = &mut self.segments[segment_id as usize];
-        if amount > seg.capacity - seg.allocated {
-            None
-        } else {
-            let result = seg.allocated;
-            seg.allocated += amount;
-            Some(result)
-        }
+        self.segments[segment_id as usize].allocate(amount)
     }
 
     fn allocate_anywhere(&mut self, amount: u32) -> (SegmentId, u32) {
         // first try the existing segments, then try allocating a new segment.
-        let allocated_len = self.segments.len() as u32;
-        for segment_id in 0..allocated_len {
-            if let Some(idx) = self.allocate(segment_id, amount) {
-                return (segment_id, idx);
+        for (segment_id, segment) in self.segments.iter_mut().enumerate() {
+            if let Some(idx) = segment.allocate(amount) {
+                return (segment_id as u32, idx);
             }
         }
 
         // Need to allocate a new segment.
 
+        let segment_id = self.segments.len() as u32;
         self.allocate_segment(amount);
         (
-            allocated_len,
-            self.allocate(allocated_len, amount)
+            segment_id,
+            self.allocate(segment_id, amount)
                 .expect("use freshly-allocated segment"),
         )
     }
--- a/capnp/src/private/arena.rs
+++ b/capnp/src/private/arena.rs
@@ -312,7 +312,7 @@ where
         self.segments.push(BuilderSegment {
             ptr: seg.0,
             capacity: seg.1,
-            allocated: 0,
+            allocated: minimum_size,
         });
     }
 
@@ -332,11 +332,7 @@ where
 
         let segment_id = self.segments.len() as u32;
         self.allocate_segment(amount);
-        (
-            segment_id,
-            self.allocate(segment_id, amount)
-                .expect("use freshly-allocated segment"),
-        )
+        (segment_id, 0)
     }
 
     fn deallocate_all(&mut self) {
--- a/capnp/src/private/arena.rs
+++ b/capnp/src/private/arena.rs
@@ -206,7 +206,7 @@ where
     }
 
     /// Allocates a new segment with capacity for at least `minimum_size` words.
-    pub fn allocate_segment(&self, minimum_size: u32) -> Result<()> {
+    pub fn allocate_segment(&self, minimum_size: u32) {
         self.inner.borrow_mut().allocate_segment(minimum_size)
     }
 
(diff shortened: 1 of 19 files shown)


dwrensha:
Wow, cool. Thanks! That `RefCell` has bothered me for a long time, but I had convinced myself somehow that removing it was infeasible.

Were you able to measure a performance improvement from this change? When I run the benchmarks in the repo, it looks like things slow down a bit.
```
$ cargo build --release -p benchmark
$ ./target/release/run_all_benchmarks ./target/release/benchmark
```

marmeladema:
I'll check the benchmarks tomorrow, thank you for pointing me in the right direction!
Are they time based? If so I am not sure I will get reproducible results on my machines. Frequency vary too much and too quickly, even with turbo boost disabled and core pinning.

dwrensha:
Yeah, the benchmarks in the repo just check the system clock before and after processing some randomized data.

marmeladema:
Not sure how reproducible my results are, but I tried running the benchmarks on Linux using the performance governor, with highest process priority on a mostly idle core.
The new code is consistently faster:
```patch
--- bench.master.log	2023-01-14 15:13:34.449673954 +0100
+++ bench.refcell.log	2023-01-14 15:10:17.616132232 +0100
@@ -1,71 +1,71 @@
 running carsales with 10000 iterations
 object none reuse
-0.274780196
+0.262815291
 
 object none no-reuse
-0.314452609
+0.290588508
 
 bytes none reuse
-0.323921303
+0.303009961
 
 bytes none no-reuse
-0.367289562
+0.337183638
 
 bytes packed reuse
-0.522327291
+0.483596268
 
 bytes packed no-reuse
-0.583030896
+0.543108862
 
 pipe none reuse
 exit status: 0
-0.379240224
+0.359025021
 
 pipe none no-reuse
 exit status: 0
-0.415696975
+0.391269543
 
 pipe packed reuse
 exit status: 0
-0.591989011
+0.561077035
 
 pipe packed no-reuse
 exit status: 0
-0.657493158
+0.633822245
 
 running catrank with 1000 iterations
 object none no-reuse
-0.650825717
+0.630079401
 
 bytes none no-reuse
-0.705408293
+0.674896325
 
 bytes packed no-reuse
-0.943492149
+0.844120619
 
 pipe none no-reuse
 exit status: 0
-0.781824298
+0.73111535
 
 pipe packed no-reuse
 exit status: 0
-0.971964107
+0.939981865
 
 running eval with 200000 iterations
 object none no-reuse
-0.704358592
+0.633409412
 
 bytes none no-reuse
-0.841810654
+0.786843015
 
 bytes packed no-reuse
-1.291954872
+1.207431156
 
 pipe none no-reuse
 exit status: 0
-1.626859737
+1.523820484
 
 pipe packed no-reuse
 exit status: 0
-2.016225853
+1.960986856
```

capnp/src/private/arena.rs
```diff
@@ -154,15 +155,9 @@ where
 }
 
 pub trait BuilderArena: ReaderArena {
-    // These methods all take an immutable &self because otherwise a StructBuilder<'a>
-    // would need a `&'a mut BuilderArena` and `StructBuilder::borrow()` would
-    // have lifetime issues. (If `'a: 'b`, then a `&'a (BuilderArena + 'a)` can be
-    // converted to a `&'b (BuilderArena + 'b)`, but a `&'a mut (BuilderArena + 'a)`
-    // *cannot* be converted to a `&'b mut (BuilderArena + 'b)`. See some discussion here:
-    // https://botbot.me/mozilla/rust/2017-01-31/?msg=80228117&page=19 .)
```
dwrensha, line 162: Ah, so this comment captures the reason why I thought that the `RefCell` was necessary. (I updated it for `dyn` and `reborrow` here: https://github.com/capnproto/capnproto-rust/commit/5ad7849382d86e6ed8043a9b7ba3c52fbc631304.)

I wonder what changed so that this is now possible? Or maybe I was mistaken all along?

dwrensha: Wow, this failed in Rust 1.62: https://godbolt.org/z/4j354zad3
and started working in Rust 1.63: https://godbolt.org/z/3hv5x3vrK

I'm a little bit worried that this means that the borrowchecker is now unsound! But if it's not, then this is good news for us, because it will let us make these methods take `&mut self` and therefore let us remove the `RefCell`.

marmeladema: I think this is because NLL has been fully stabilized in 1.63.0: https://blog.rust-lang.org/2022/08/05/nll-by-default.html

marmeladema: Looking at the code, I don't see any obvious reasons it should not compile or can be unsound.

marmeladema: I can probably ask t-types team on zulip for a confirmation?

dwrensha: Aha. Here's a superficially similar thing that (thankfully) still does not typecheck:
```rust
struct Foo<'a> {
    buf: &'a mut [u8],
}

struct Bar<'a> {
    inner : &'a mut Foo<'a>,
}

impl <'a> Bar <'a> {
    fn reborrow<'b>(&'b mut self) -> Bar<'b>  {
        Bar { inner: self.inner }
    }
}

pub fn main() {}
```

Allowing this would be unsound, because then we could do:
```rust
pub fn main() {
   let mut v1 : Vec<u8> = vec![1,2,3];
   let mut foo1 = Foo { buf: &mut v1 };
   let mut bar = Bar { inner: &mut foo1 };
   {
        let mut v2 : Vec<u8> = vec![4,5];
        let mut foo2 = Foo { buf: &mut v2 };
        *bar.reborrow().inner = foo2;
   }
   println!("{}", bar.inner.buf[0]); // accesses v2, which is out of scope!
}
```

The problem is that we can *assign* a value to `Bar::inner` that does not live long enough.

In our current case, however, we are dealing with trait objects (i.e. `dyn`), and those do *not* allow assignment through `&mut` references (because the size is unknown). So I don't see any obvious unsoundness.

dwrensha: That's to say, it's perhaps understandable that rustc pre-NLL was too conservative in this case.

dwrensha: Ah, and according to the referece (https://doc.rust-lang.org/reference/subtyping.html#variance),

`dyn Trait<T> + 'a` is "covariant" in `'a`. Pre-NLL it was "invariant", if I understand correctly.

dwrensha: Actually, I can get my little example to typecheck on an even older rustc (e.g. 1.30) if I avoid the `..*self` syntax:
https://godbolt.org/z/8EGx4v3zK

dwrensha: posted on zulip: https://rust-lang.zulipchat.com/#narrow/stream/144729-t-types/topic/subtype.20covariance.20with.20.60.26'a.20mut.20.28dyn.20R.20.2B.20'a.29.60/near/321516958

marmeladema: If I understand correctly, 1.63.0 just made it work with the `..*self` syntax, but it was working before using regular field assignment. So arguably, we could have introduced such code before 1.63.0 and get rid of `RefCell` already.

So answering your first question from zulip:

> Can I rely on this continuing to typecheck? I'd like to migrate a bunch of my code to depend on this, because I had been working around it previously.

Unless there is something inherently unsound about accepting this pattern; I fail to see why the compiler would suddenly reject it. That code has been working for years and such pattern is probably used in various places; so rejecting it would be tricky.

Given all of this, do you see any blocker that would prevent merging this?

dwrensha: I want to convince myself it's not unsound first. :)

I think this PR should be considered a breaking change and that we should do a version bump soon including it and some of the other stuff that's been waiting on a version bump.

dwrensha:
Superceded by #365.

PR #364, opened 2023-01-14
Use `SmallVec` in builder arena to avoid heap allocation for single-segment case

marmeladema:


dwrensha:
Does this actually help performance? Last time I tried it seemed to make things a bit slower.

marmeladema:
I never meant to close this 😑

marmeladema:
I haven't run the benchmarks yet, and to be honest, I don't care much about this because I have another change I want to propose on top of the refactor build arena PR where I make a builder re-usable.

dwrensha:
> where I make a builder re-usable.

What kind of builder? A `message::Builder`? Can you explain more about your plan?

Does the existing `message::ScratchSpaceHeapAllocator` work for your use case?

marmeladema:
The allocator is only part of the solution. I want to re-use the vec of segments in the builder arena.

PR #365, opened 2023-01-16
Remove `RefCell` from builder arena

dwrensha:
This is #363 without the changes from #362.


dwrensha:
This seems to produce a noticeable speedup on the serdebench benchmarks: https://github.com/llogiq/serdebench/pull/19

PR #367, opened 2023-01-17
suppress 'clippy::needless_update' lint

danieleades:


danieleades:
hmm rerunning the codegen seems to have had a greater impact than just my changes. I'm guessing this wasn't done after a change to the codegen in a different PR?

Seems wrong that automatically generated files have to be manually generated and committed at all. Is it possible to fix this, or is there some circular dependency that prevents it?

danieleades:
given the volume of `allow` calls, i'm almost attempted to just have one global allow at the top of the module, rather than at the call site. What do you think @dwrensha ?

dwrensha:
See #369 and https://github.com/capnproto/capnproto-rust/commit/3342c4344f48b1ee6b9195da495bcbeba50fb5ee

PR #369, opened 2023-01-17
fix clippy warning

zh-jq:


danieleades:
this looks to be a better solution than #367

dwrensha:
Thanks!

danieleades:
The generated files need to be regenerated though

PR #371, opened 2023-01-17
move msrv from clippy config to Cargo.toml

danieleades:
clippy now supports parsing the 'rust version' from cargo.toml

setting it in Cargo.toml has other benefits, such as allowing compilation to exit gracefully if the compiler version doesn't meet the `rust-version` minimum

this patterns sets the same value for all members of the workspace.


danieleades:
> `rust-version` got added to `capnp/Cargo.toml` in #356 If we add it to the workspace Cargo.toml, should we keep or remove that one? If it's in the workspace Cargo.toml, does it get picked up by packaged crates?

ah good catch. So in this case the version set in the root is inherited by all the members of the workspace, but they can override it locally.
so the field should be removed from capnp/Cargo.toml in favour of the root.

i'll update

dwrensha:
It looks like maybe the workspace-level msrv does not get picked up by packaging?

```
$ cargo package -p capnp
   Packaging capnp v0.16.0 (/home/dwrensha/src/capnproto-rust/capnp)
   Verifying capnp v0.16.0 (/home/dwrensha/src/capnproto-rust/capnp)
    Updating crates.io index
   Compiling capnp v0.16.0 (/home/dwrensha/src/capnproto-rust/target/package/capnp-0.16.0)
    Finished dev [unoptimized + debuginfo] target(s) in 1.19s
    Packaged 40 files, 418.9KiB (70.7KiB compressed)
dwrensha@hafnium:~/src/capnproto-rust$ cat target/package/capnp-0.16.0/Cargo.toml | grep rust-version
[not found]
```

I suppose that means we need it in both places?

danieleades:
Looks like I misunderstood how the inheritance works. The docs here make it more clear - https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

you need to opt in per-key inside the member package

dwrensha:
Thanks!

PR #376, opened 2023-01-30
avoid ambiguous associated item in `TryFrom` implementations

hunts:
This was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!

See also: https://github.com/rust-lang/rust/issues/57644

Without this change, latest versions of cargo would report this:

```
error: ambiguous associated item
   --> src/log_capnp.rs:213:69
    |
213 |             fn try_from(value: u16) -> ::core::result::Result<Self, Self::Error> {
    |                                                                     ^^^^^^^^^^^ help: use fully-qualified syntax: `<Level as TryFrom>::Error`
    |
```

### DIFF
--- a/capnpc/src/codegen.rs
+++ b/capnpc/src/codegen.rs
@@ -1937,7 +1937,7 @@ fn generate_node(
                 ))),
                 Indent(Box::new(Branch(vec![
                     Line(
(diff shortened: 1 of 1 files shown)


dwrensha:
Thanks! Fixed in https://github.com/capnproto/capnproto-rust/commit/59f2ad314bd14a868044ed960c4442677338061a. (Your version didn't quite work.)

PR #384, opened 2023-03-13
Improve miri testing

marmeladema:


capnp/src/serialize/no_alloc_slice_segments.rs
```diff
@@ -330,24 +329,37 @@ mod tests {
...
+    struct Aligned([u8; 8]);
+
     #[cfg(feature = "unaligned")]
     #[test]
     fn test_verify_alignment_unaligned_mode() {
         // To run this test do
         // `% cargo test --features unaligned`
 
+        // make sure there is no padding
+        assert_eq!(std::mem::size_of::<Aligned>(), 8);
```
dwrensha, line 342: Please use `core` instead of `std`, so that this works in a no_std environment.

marmeladema: Done

PR #386, opened 2023-04-06
fix clippy 1.70 warning

zh-jq:


zh-jq:
MSRV is 1.65, so is there any need to still support rust 2015 and 2018?

dwrensha:
Projects using rustc 1.65+ can still use older editions. `cargo fix` cannot always automatically upgrade a project to a newer edition. I imagine there are many projects still using older editions that have not bothered to upgrade yet.

How hard is it for you to suppress the clippy warnings on your end?

zh-jq:
close this as the clippy warning is not used in 1.70

Issue #145, opened 2019-09-13
Please improve getting started instructions -- `include!`


xpe:
The current getting started text from https://crates.io/crates/capnp-rpc would benefit from additional clarity in the following section...

```
and you can include the generated code in your project like this:

pub mod foo_capnp {
  include!(concat!(env!("OUT_DIR"), "/foo_capnp.rs"));
}
```

My questions are:
* Where exactly does `mod foo_capnp` belong?
* How is the path for "foo_capnp.rs" generated? I'm digging around looking for it in my 'target' directory.

dwrensha:
There's some documentation about this here: https://github.com/capnproto/capnproto-rust/blob/96f2f7894fbbf55e741c39b183a824e27ddbcc3e/capnpc/src/lib.rs

I agree that more/better documentation would be good. Perhaps the best place to discuss how to `include!()` would be here: https://github.com/capnproto/capnproto-rust/blob/96f2f7894fbbf55e741c39b183a824e27ddbcc3e/capnp/src/lib.rs

as then it would be at the top of the first page that people see when clicking on the documentation link in the readme: https://docs.capnproto-rust.org/capnp/

dwrensha:
#573 adds a macro that simplifies importing generated code.

dwrensha:
Added a few sentences to the README to answer your original questions: https://github.com/capnproto/capnproto-rust/commit/58977b7ae792c413df21a75f982a406cc56a31eb

PR #387, opened 2023-04-10
Improve getting started instructions

danielzfranklin:
I removed the redundant `build = "build.rs"`, added the necessary dependency on `capnp`, and provided examples of including the generated files.

I started using capnproto five minutes ago, this is just a guess at the right getting started instructions. I thought I'd throw together a PR with an incremental improvements because the original issue about better instructions is four years old.

Addresses #145


dwrensha:
Thanks!

PR #389, opened 2023-04-11
Add WeakCapabilityServerSet

### LINKED ISSUE #388: Using `CapabilityServerSet` prevents Server objects from being dropped at the appropriate time
When I use a `CapabilityServerSet` to create clients for Server objects, those Server objects are no longer dropped when the remote connection drops them. These Server objects are kept alive until the `CapabilityServerSet` itself is dropped. 

The behavior I want is for the reference to the Server object to be removed from the `CapabilityServerSet` when it is dropped by the remote connection, so that the Server object may be dropped immediately.

Is it possible to achieve this?

canarysnort01:
Allow Server objects to be dropped when the remote client drops them. Also adds a `gc` method to reclaim memory used for Server objects that have been dropped.

Fixes #388

### DIFF
--- a/capnp-rpc/src/lib.rs
+++ b/capnp-rpc/src/lib.rs
@@ -65,7 +65,7 @@ use futures::channel::oneshot;
 use futures::{Future, FutureExt, TryFutureExt};
 use std::cell::RefCell;
 use std::pin::Pin;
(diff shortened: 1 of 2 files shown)


dwrensha:
Thanks!

PR #403, opened 2023-06-04
Add annotation for getting field as Rust Option

### LINKED ISSUE #249: Mandatory value accessors
Hello and thank you for your work!

I am working on a project that uses a capnproto schema where most of the values are mandatory when reading a message.
I got bitten multiple times by the fact that a `reader.get_field()` method for a `field @0 :Text` member will return an empty string when `field` is not specified in the message.
Indeed I am now using `reader.has_field()` with the following pattern:
```Rust
let key = if reader.has_key() {
    reader.get_key()?
} else {
    return Err(::capnp::Error::failed("Missing `key` parameter".to_owned()));
};
```

This is rather cumbersome to do everywhere and I wonder if the API could be improved.
Would it make sense to add a new error type that could be called `MissingValue` (I don't care much about the actual name) and add a new `try_get_<name>` methods in readers to retrieve a value if it exist and fail otherwise without ever returning a default value if one hasn't been specified in the schema.

The sounds fine in principle but `get_<name>` reader methods already return a `Result`which means both methods would have the same prototype, only proper documentation would differentiate them:
* `get_<name>` methods will return a value, and sometime a default one for types that have an implicitly defined value like `Text` or integer fields
*  `try_get_<name>` methods will only return a value when it is present in the message

Thank you for taking the time to read me. Maybe my solution is totally off but the ergonomic issues are real though :)

danielzfranklin:
Given

```capnp
struct Test {
    field @0 :Text $Rust.option;
}
```

you get getters like so

```rust
assert_eq!(struct_with.get_field(), Some("foo"));
assert_eq!(struct_without.get_field(), None));
```

The setters are unchanged to match the Rust convention.

Fixes #249.

### DIFF
--- a/capnp/src/private/layout.rs
+++ b/capnp/src/private/layout.rs
@@ -3479,6 +3479,11 @@ impl<'a> StructReader<'a> {
         }
     }
 
+    #[inline]
+    pub fn is_pointer_field_null(&self, ptr_index: WirePointerCount) -> bool {
+        unsafe { (*self.pointers.add(ptr_index)).is_null() }
+    }
+
     pub fn total_size(&self) -> Result<MessageSize> {
         let mut result = MessageSize {
             word_count: u64::from(wire_helpers::round_bits_up_to_words(u64::from(
--- a/capnp/src/private/layout.rs
+++ b/capnp/src/private/layout.rs
@@ -3481,7 +3481,11 @@ impl<'a> StructReader<'a> {
 
     #[inline]
     pub fn is_pointer_field_null(&self, ptr_index: WirePointerCount) -> bool {
-        unsafe { (*self.pointers.add(ptr_index)).is_null() }
+        if ptr_index < self.pointer_count as WirePointerCount {
+            unsafe { (*self.pointers.add(ptr_index)).is_null() }
+        } else {
+            true
+        }
     }
 
     pub fn total_size(&self) -> Result<MessageSize> {
--- a/capnpc/rust.capnp
+++ b/capnpc/rust.capnp
@@ -26,3 +26,20 @@ annotation parentModule @0xabee386cd1450364 (file) :Text;
 #      }
 #    }
 #  }
+
+annotation getOption @0xabfef22c4ee1964e (field) :Void;
+# Whether the generated code should return an Option. Supported on Text, Data,
+# List, and struct fields.
+#
+# Given
+#
+#     struct Test {
+#         field @0 :Text $Rust.getOption;
+#     }
+#
+# you get getters like so
+#
+#     assert_eq!(struct_with.get_field(), Some("foo"));
+#     assert_eq!(struct_without.get_field(), None));
+#
+# The setters are unchanged to match the Rust convention.
--- a/capnpc/rust.capnp
+++ b/capnpc/rust.capnp
@@ -28,8 +28,8 @@ annotation parentModule @0xabee386cd1450364 (file) :Text;
 #  }
 
 annotation getOption @0xabfef22c4ee1964e (field) :Void;
-# Whether the generated code should return an Option. Supported on Text, Data,
-# List, and struct fields.
+# Whether the generated code should return an Option. Supported on pointer
+# fields.
 #
 # Given
 #
--- a/capnpc/rust.capnp
+++ b/capnpc/rust.capnp
@@ -27,14 +27,18 @@ annotation parentModule @0xabee386cd1450364 (file) :Text;
 #    }
 #  }
 
-annotation getOption @0xabfef22c4ee1964e (field) :Void;
-# Whether the generated code should return an Option. Supported on pointer
-# fields.
+annotation option @0xabfef22c4ee1964e (field) :Void;
+# Make the generated getters return Option<T> instead of T. Supported on
+# pointer types (e.g. structs, lists, and blobs).
+#
+# Capnp pointer types are nullable. Normally get_field will return the default
+# value if the field isn't set. With this annotation you get Some(...) when
+# the field is set and None when it isn't.
 #
 # Given
 #
 #     struct Test {
-#         field @0 :Text $Rust.getOption;
+#         field @0 :Text $Rust.option;
 #     }
 #
 # you get getters like so
@@ -43,3 +47,5 @@ annotation getOption @0xabfef22c4ee1964e (field) :Void;
 #     assert_eq!(struct_without.get_field(), None));
 #
 # The setters are unchanged to match the Rust convention.
+#
+# Note: Support for this annotation on interfaces isn't implemented yet.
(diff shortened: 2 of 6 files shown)


dwrensha:
Interesting!

> The setters are unchanged to match the Rust convention.

What Rust convention you are you referring to here? Are there some methods in `std` that follow this pattern?

danielzfranklin:
> What Rust convention you are you referring to here? Are there some methods in std that follow this pattern?

I'm referring to the convention that when you have a builder with an optional field you don't have the setter take an option. Instead the user just doesn't call the setter if they don't want it set.

One example would be [`std::thread::Builder::name`](https://doc.rust-lang.org/std/thread/struct.Builder.html#method.name)

capnpc/rust.capnp
```diff
@@ -26,3 +26,20 @@ annotation parentModule @0xabee386cd1450364 (file) :Text;
 #      }
 #    }
 #  }
+
+annotation getOption @0xabfef22c4ee1964e (field) :Void;
+# Whether the generated code should return an Option. Supported on Text, Data,
+# List, and struct fields.
```
zenhack, line 32: Is there any reason for this to not work on any pointer type?

zenhack (commented):
I've been thinking of adding something similar for the Haskell implementation -- maybe I will steal the node id.

nit: I'm tempted to suggest a more declarative name, like `nullable` or `optional`.

capnpc/rust.capnp
danielzfranklin: I forgot :). Went back to the capnp docs and I think I'm just missing AnyPointer and interfaces.

Just implemented AnyPointer support.

I just spent about an hour hitting my head against a wall trying to understand capnp interfaces, which I've never used before. I couldn't figure out how to do something as simple as set a field in a message to an empty interface. There are some places in the current code that look buggy (for ex I think [this](https://github.com/danielzfranklin/capnproto-rust/blob/f82031359ebf4e8824d09d416f97d85890de74b1/capnpc/src/codegen.rs#LL664C40-L664C41) breaks interfaces nested in things). I don't see any examples or tests for interfaces in this repo.

Do you know any examples of code using interfaces I could read? I'm considering leaving them out of this PR.

danielzfranklin:
> nit: I'm tempted to suggest a more declarative name, like `nullable` or `optional`.

My thinking is this modifies the way you read the data but your data semantically stays the same.

With something like the following I worry people would interpret it to mean that `fieldA` must be set but `fieldB` is optional.

```capnp
struct Foo {
   fieldA @1 :Text;
   fieldB @2 :Text $Rust.optional;
}
```

capnp/src/private/layout.rs
```diff
@@ -3479,6 +3479,11 @@ impl<'a> StructReader<'a> {
         }
     }
 
+    #[inline]
+    pub fn is_pointer_field_null(&self, ptr_index: WirePointerCount) -> bool {
```
dwrensha, line 3483: What if `ptr_index` is greater than or equal to `self.pointer_count`? That can happen if the writer of the message is using an old version of the schema that has fewer fields defined.

danielzfranklin: I was assuming that the check had to be somewhere else because otherwise I don't see how the rest of this file works.

I copied the implementation of `StructReader::is_pointer_field_null`.

Doesn't `StructBuilder::get_pointer_field(...).is_null()` also ultimately boil down to the same an unchecked addition?

```rust
impl<'a> StructBuilder<'a> {
    // ...
    pub fn get_pointer_field(self, ptr_index: WirePointerCount) -> PointerBuilder<'a> {
        PointerBuilder {
            arena: self.arena,
            segment_id: self.segment_id,
            cap_table: self.cap_table,
            pointer: unsafe { self.pointers.add(ptr_index) },
        }
    }
    // ...
}

// ...
    
impl<'a> PointerBuilder<'a> {
     // ...
    #[inline]
    pub fn is_null(&self) -> bool {
        unsafe { (*self.pointer).is_null() }
    }
    // ...
}
 ```

Comment (dwrensha) [reply capnp/src/private/layout.rs:3483]: Builders are guaranteed to have the expected number of pointers. Readers are not.

**Comment by dwrensha** [comment]
I wonder whether the `Builder` getter methods should also return an `Option`, for symmetry.

Also, it occurs to me that our generated code does not include a way to clear a pointer field. I think that in capnproto-c++ the `disown` methods do that. Maybe we should just add `clear` methods for all pointer fields. (I don't intend to support the capnproto-c++ notion of "orphans" any time soon.) (Such a thing probably does not belong in this pull request, but I do want to make sure we have a plausible plan going forward.)

File: capnpc/src/codegen.rs [hunk 3]
```diff
@@ -706,80 +742,109 @@ pub fn getter_text(
...
                 _ => return Err(Error::failed("default value was of wrong type".to_string())),
             };
+
+            let getter_code = if should_get_option && is_reader {
+                Branch(vec![
+                    Line(format!(
+                        "if self.{member}.is_pointer_field_null({offset}) {{"
+                    )),
+                    Indent(Box::new(Line(
+                        if is_fallible { "Ok(None)" } else { "None" }.to_string(),
```
dwrensha, line 834: `::core::option::None` to minimize chance of name collision.

capnpc/src/codegen.rs
```diff
@@ -706,80 +742,109 @@ pub fn getter_text(
...
+                    ))),
+                    Line("} else {".to_string()),
+                    Indent(Box::new(Line(if is_fallible {
+                        format!("{getter_fragment}.map(Some)")
+                    } else {
+                        format!("Some({getter_fragment})")
```
dwrensha, line 840: `::core::option::Option::Some`

danielzfranklin:
> I wonder whether the `Builder` getter methods should also return an `Option`, for symmetry.

I don't have a strong opinion here. I've yet to find myself using the builder getters. I wasn't sure if they're actually symmetrical. 

> Also, it occurs to me that our generated code does not include a way to clear a pointer field. I think that in capnproto-c++ the `disown` methods do that. Maybe we should just add `clear` methods for all pointer fields. (I don't intend to support the capnproto-c++ notion of "orphans" any time soon.) (Such a thing probably does not belong in this pull request, but I do want to make sure we have a plausible plan going forward.)

That makes sense. I don't see how this PR would interfere with a clear method.

dwrensha:
> I don't have a strong opinion here. I've yet to find myself using the builder getters. I wasn't sure if they're actually symmetrical.

Ok, I think it makes most sense if the builders match the readers on this.

dwrensha:
Of the possible names brought up so far, I think I like `$Rust.optional` the best.

I think `getOption` is a bit ambiguous, and feels a bit too much like a command -- is the "get" an action that I need to take as a user? Or maybe it's something that automatically happens in the background? Or maybe `get` is an adjective modifying `Option`?

danielzfranklin:
What about simply `$Rust.option`?

I agree on further thought that `get` is too much like a command, but I feel like the adjective `optional` might imply this affects validation.

dwrensha:
Yeah, `$Rust.option` works for me.

PR #405, opened 2023-06-05
Opt-in native getters and setters for "newtypes"/"value types"

### LINKED ISSUE #245: Support newtypes
I'm researching capnproto with intention to use it soon. From what I can see there is no support for newtypes. I believe it'd be useful.

Examples of some newtypes I have in mind:

```rust
struct Amount(u64);

struct Hash([u8; 32]);
```

I imagine there could be a way to annotate fields to use newtypes with specific validation function. `Reader` would call this function and return the newtype or error:

```rust
// defined by consumer or the crate providing Amount
fn validate_amount(buf: &[Word]) -> Result<&Amount> {
    if buf.len() != 1 {
        return Err(Error { kind: ErrorKind::Fail, message: "length is not 1".to_owned() });
    }
    if cast_word_to_u64(buf[0]) > LIMIT {
        return Err(Error { kind: ErrorKind::Fail, message: "value over limit".to_owned() });
    }
    Ok(unsafe { &*(buf.as_ptr() as *const Amount) })
}

// generated code
impl Reader<'_> {
    pub fn get_amount(&self) -> Result<&Amount> {
        // validates pointers etc
        let words = self.internal_get_amount();
        validate_amount(words)
    }
}
```

The consumers would just call `get_amount()` and get the validated newtype out of the box.

Similarly, serialization could be implemented.

danielzfranklin:
My goal is to incrementally make the generated code easier to use without getting bogged down in something ambitious (#136). This is intended for simple types that nearly primitive values.

To opt-in you add an annotation and implement a trait

```capnp
struct User {
    id @0 :Uuid;
}

struct Uuid $Rust.valueType {
  d1 @0 :UInt32;
  d2 @1 :UInt16;
  d3 @2 :UInt16;
  d4 @3 :UInt64;
}
```

```rust
use uuid::Uuid;

impl capnp::traits::ValueType for account_capnp::uuid::Owned {
    type Type = Uuid;

    fn read(reader: Self::Reader<'_>) -> capnp::Result<Self::Type> {
        let d4 = reader.get_d4().to_be_bytes();
        Ok(Uuid::from_fields(reader.get_d1(), reader.get_d2(), reader.get_d3(), &d4))
    }

    fn build(mut builder: Self::Builder<'_>, value: Self::Type) -> capnp::Result<()> {
        let (d1, d2, d3, d4) = uuid.as_fields();
        let d4 = u64::from_be_bytes(*d4);
        builder.set_d1(d1);
        builder.set_d2(d2);
        builder.set_d3(d3);
        builder.set_d4(d4);
        Ok(())
    }
}
```

```rust
let message = capnp::message::Builder::new_default();
let root = message.init_root::<account_capnp::user::Builder>();

let admin_id = Uuid::from_str("a1b2c3d4-e5f6-7a8b-9c0d-1e2f3a4b5c6d").unwrap();
root.set_id(admin_id);
assert_eq!(root.reborrow_as_reader().get_id(), admin_id);
```

Fixes #245


danielzfranklin:
I've ended up thinking this isn't worthwhile

PR #406, opened 2023-06-06
Get miri passing again and add to CI

danielzfranklin:
The only substantive change is [using a raw pointer instead of a reference](https://github.com/capnproto/capnproto-rust/pull/406/commits/a4874defa8b6a3616d0b0cbf0574289ff4387238). This is a change to a public function in capnp::private so I don't think it's a breaking change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,39 @@ jobs:
     - name: Run benchmark
       run: ./target/release/run_all_benchmarks target/release/benchmark
 
+  miri:
+    name: "Miri"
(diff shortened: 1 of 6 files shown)


dwrensha:
Is miri reliable enough to run in CI nowadays? Last I checked, it was broken on a high percentage of nightly rustc releases.

danielzfranklin:
The configuration I used will test on the latest nightly where miri works.

dwrensha:
OK, I'm willing to try it. We can disable it later if it's unreliable.

PR #407, opened 2023-06-09
Implement json serialization/deserialization with the reflection api (WIP)

danielzfranklin:
I really like the new reflection API. Implementing JSON ser/de has been quite straightforward.

At this point the remaining work basically just unions. I'm opening this draft PR to avoid accidental duplication of work.

Using the API has exposed a bug and a few api ergonomics additions. I'm happy to break those out separately if you prefer.


dwrensha:
> Using the API has exposed a bug and a few api ergonomics additions. I'm happy to break those out separately if you prefer.

Yes, if there are fixes that are independent of the json stuff, then I would prefer to see those in separate PRs.

dwrensha:
Did you consider splitting out a separate crate, maybe `capnp_json`, rather than using a Cargo feature? That might be a cleaner way to keep things organized.

danielzfranklin:
> Did you consider splitting out a separate crate, maybe `capnp_json`, rather than using a Cargo feature? That might be a cleaner way to keep things organized.

That's a good idea. Do you think that crate should live in this repository?

dwrensha:
I was thinking it would still live in this repository.

Issue #411, opened 2023-06-11
get_annotation_types breaks on non-builtin external types, despite comment in codegen.rs saying otherwise


danielzfranklin:
The comment says "Avoid referring to the annotation in the generated code, so that users can import annotation schemas like `c++.capnp` or `rust.capnp` without needing to generate code for them, as long as the annotations are not generic." But the generated code does refer to annotation types.

This breaks if the annotation has a non-builtin type defined externally.

https://github.com/capnproto/capnproto-rust/blob/39a524888ffcf0dfaf9e58c234c9491e01bd7cba/capnpc/src/codegen.rs#L1662

For example

```capnp
using Json = import "/capnp/compat/json.capnp";

struct SimpleNamedUnion {
  value :union $Json.discriminator(name = "renamed-value", valueName = "type") {
    unset @0 :Void;
    variant @1 :UInt8;
  }
}
```

generates the following, which won't compile

```rust
    pub fn get_annotation_types(child_index: Option<u16>, index: u32) -> ::capnp::introspect::Type {
        match (child_index, index) {
          (Some(0), 0) => <crate::json_capnp::discriminator_options::Owned as ::capnp::introspect::Introspect>::introspect(),
          _ => panic!("invalid annotation indices ({:?}, {}) ", child_index, index),
        }
    }
```
The simplest solution I can think of that doesn't require users to do anything but works for all annotations would be for us to codegen the necessary the annotation types in a private module in each file generated. That adds a little duplication to save users from having to include_str! a new file, which I think is probably worthwhile. We can reduce the bloat by passing a flag to codegen_node telling it to skip generating the Reader & Builder.

dwrensha:
This potentially relates to #233. You could imagine that the `capnp_json` crate might be the canonical place for the `json.capnp` schema to live, but then downstream crates will want a way to import that schema.

danielzfranklin:
> This potentially relates to #233.

I think annotations are only used in the `.capnp` files of downstream consumers, never in their rust code. `capnp_json` is wholly responsible for the rust code which reads JSON annotations. So it's not important for consumers to get the generated rust code, just the `.capnp` file. And copy-pasting that is not a huge deal.

We also want to be able to ignore annotations we don't care about. I shouldn't need to add a `capnp_cpp` crate just to be able to compile a capnproto schema that is shared between Rust and C++ projects and uses C++ annotations with non-builtin types.

[...]

Issue #233, opened 2021-09-19
add support for cross-crate schema imports


dwrensha:
See https://groups.google.com/g/sandstorm-dev/c/hdTrPu6gndY/m/heRh-ns1AgAJ

Currently, if someone wants to define schema files that depend on the sandstorm schema files, they cannot use the [sandstorm-rust](https://github.com/dwrensha/sandstorm-rust) crate for it. To make that work, we need:

1. a way for the downstream crate to find the .capnp files of the upstream crate. Here's an cargo issue I opened about that: https://github.com/rust-lang/cargo/issues/3544
2. a way for capnpc-rust to know the crate name to associate with a given schema. One way to do that might be to add new annotation in rust.capnp: https://github.com/capnproto/capnproto-rust/blob/0f7abd17ba653ea2130185ca20ebb953dc729c5e/capnpc/rust.capnp

zenhack:
Re: (1), it looks like there's a consensus on the solution and this is just waiting on someone to do the work.

Re: (2), the go implementation does something similar, and it's worked out adequately.

griff:
I have gotten around the problem by having a checked out version of sandstorm-rust for calling capnpc so the schema files could be found.

And the crate problem can be solved with `use`. If my crate uses `util.capnp` it just needs `use sandstorm::util_capnp;` at the top level to work.

I actually did some work on DEP_FOO_KEY in cargo and got something hacked together but when the cargo tests failed for some reason I but it down and haven't looked at it since.

PR #412, opened 2023-06-12
Add CompilerCommand::link_override

danielzfranklin:
Relates to #411 and #233 

```capnp
annotation imports @0xc3b9fe42a83105cd (file) :List(Import);
# Allows specifying that the generated code for an import is located in another
# crate.
#
# You only need this if your code uses imported types whose generated code is
# in another crate. You can only use this annotation once in all the files
# compiled together.
#
#    using Json = import "/capnp/compat/json.capnp";
#
#    $Rust.imports([
#        (path = "/capnp/compat/json.capnp", crate = "capnp_json")
#    ]);

struct Import {
    path @0 :Text;
    crate @1 :Text;
}
```

I'm intending this as a stopgap until cargo supports non-linking DEP environment variables. It allows people to use generated code in external crates without us needing to add an annotation to the external `.capnp` file. This is important because without the change to cargo it would be awkward to require users use a modified `json.capnp`.

Unfortunately capnproto doesn't support annotations on import statements.


capnpc/rust.capnp
```diff
@@ -27,6 +27,32 @@ annotation parentModule @0xabee386cd1450364 (file) :Text;
 #    }
 #  }
 
+annotation imports @0xc3b9fe42a83105cd (file) :List(Import);
+# Allows specifying that the generated code for an import is located in another
+# crate.
+#
+# You only need this if your code uses imported types whose generated code is
+# in another crate. You can only use this annotation once in all the files
+# compiled together.
```
dwrensha, line 36: Did you consider making this an option to `capnp::CompilerCommand` rather than an annotation? That might make some of this a bit cleaner.

capnpc/rust.capnp
```diff
@@ -27,6 +27,32 @@ annotation parentModule @0xabee386cd1450364 (file) :Text;
...
+#
+#    using Json = import "/capnp/compat/json.capnp";
+#
+#    $Rust.imports([
+#        (path = "/capnp/compat/json.capnp", crate = "capnp_json")
```
dwrensha, line 41: It could make sense for the key to be the 64-bit capnp ID of the file, rather than the string `path`. That seems to me like it would maybe be a bit more robust. I'm not sure which alternative would be more user friendly.

regenerate-rust-annotations.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
+
+set -e
+set -x
+
+cargo build -p capnpc
+capnp compile -otarget/debug/capnpc-rust-bootstrap:capnp/src capnpc/rust.capnp --src-prefix capnpc/
```
dwrensha, line 7: What's the reasoning behind adding a new script here, rather than using the existing `regenerate-capnp-schema-code.sh`?

capnpc/rust.capnp
danielzfranklin: Great idea!

regenerate-rust-annotations.sh
danielzfranklin: Just assumed that would be how you wanted it based on the filename and the existing pattern

danielzfranklin: This is irrelevant because I no longer need the annotation type

danielzfranklin:
The new api is like so

```rust
fn main() {
    capnpc::CompilerCommand::new()
        .link_override(0xe6f94f52f7be8fe2, "external_crate")
        .file("test.capnp")
        .run()
        .unwrap();
}
```

I chose to call it `link_override` instead of `link` because I think that'll be clearer when we also have #233 .

I think "link" is a better word here than "import". "use" or "extern" might also make sense to mirror the rust syntax.

danielzfranklin:
Agree "link" is wrong. What about `.crate_provides("capnp_json", [0x8ef99297a43a5e34])`.

dwrensha:
Works for me.

capnpc/src/codegen.rs
```diff
@@ -240,8 +257,15 @@ impl<'a> GeneratorContext<'a> {
                     "{}_capnp",
                     path_to_stem_string(importpath)?.replace('-', "_")
                 );
+                let parent_module_scope = if let Some(krate) = crates_provide.get(&import.get_id())
```
dwrensha, line 260: What should happen if a file in `ctx.request.get_requested_files()` has this ID too? Then the user is generating code for that file, and it seems to me that the crate-local version should win.

I think that, given an import ID `X`, the precedence should be:
1. If the user explicitly requests codegen a file with ID `X`, then that locally-generated version should be used, regardless of the contents of `crates_provide_map`.
2. Otherwise, if `X` appears in `crates_provide_map`, then that version of the name should be used.
3. Otherwise, if the referred-to file has a `$Rust.crate` annotation, then the `Text` crate name in that annotation should be used.
4. Otherwise, compile-time error (or graceful fallback).

I see that you removed the `$Rust.crate` annotation from this pull request. I do think we should add that annotation. Up to you whether you want it to be in this pull request or not.

danielzfranklin: I'm tempted to say that if a user explicitly says a file should be provided by a crate and also explicitly asks to compile that file themselves we should fail with an error.

My guess is the main reason this would happen is that somebody was confused and thought they had to compile their imports. I can't think of a good reason you'd ever do this.

I agree $Rust.crate should be overridden by the build script option. I removed it because I was thinking it'd only be useful once we had the cargo feature but on second thought I think it's useful right away. I'll add it back.

dwrensha: > I'm tempted to say that if a user explicitly says a file should be provided by a crate and also explicitly asks to compile that file themselves we should fail with an error.

Yeah that makes sense. I was forgetting that they need to explicitly opt into each individual file in `crate_provides()`.

danielzfranklin: In terms of fallbacks I don't think it's that important. c++.capnp doesn't have any annotations with non built-in types, so I think it's unlikely users will have such annotations in their code and not want to parse the annotations in Rust.

If it's a field instead of an annotation I can't think of a use case where a fallback would be useful. 

Another possible API would be to fail with a helpful error message that suggests adding an explicit exclude_annotations(file_id) call.

dwrensha: I was imagining a schema C++ project, where it has a bunch of $Json annotations sprinkled in. A Rust project might not care about JSON might to want to add a dependency on the `capnp_json` crate. So I think some way to ignore the annotations is needed. However, since there's probably more discussion to be had about it, and since nobody is actually complaining about it yet, I think it does not belong in the current PR.

dwrensha:
`$Rust.crate` can be added in a latter pull reqeust.

PR #414, opened 2023-06-12
Fix rustfmt of codegen.rs and add ci check for rustfmt bailing

danielzfranklin:


dwrensha:
Wow, it's surprising that `cargo fmt` would fail silently like that.

danielzfranklin:
Yeah and it's an unstable argument. The only google results are that one discussion in 2018 about how maybe this should be the default, but I guess it's too late to break everyone's build?

PR #417, opened 2023-06-23
Replace error strings with enums

ariel-miculas:
This is useful in `no alloc` scenarios and in cases where String is not
available (such as the Linux kernel).

### DIFF
--- a/capnp/src/lib.rs
+++ b/capnp/src/lib.rs
@@ -63,6 +63,7 @@ pub mod text;
 pub mod text_list;
 pub mod traits;
 
+#[cfg(feature = "alloc")]
 use alloc::string::String;
 use alloc::vec::Vec;
 
@@ -134,7 +135,7 @@ impl core::ops::AddAssign for MessageSize {
 }
 
 /// An enum value or union discriminant that was not found among those defined in a schema.
-#[derive(PartialEq, Clone, Copy, Debug)]
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
 pub struct NotInSchema(pub u16);
 
 impl ::core::fmt::Display for NotInSchema {
@@ -168,15 +169,17 @@ pub struct Error {
     /// should read only this field in making its decision.
     pub kind: ErrorKind,
 
-    /// Human-readable failure description.
-    pub description: String,
+    /// Extra context about error
+    #[cfg(feature = "alloc")]
+    pub extra: String,
 }
 
 /// The general nature of an error. The purpose of this enum is not to describe the error itself,
 /// but rather to describe how the client might want to respond to the error.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum ErrorKind {
-    /// Something went wrong. This is the usual error kind. It includes decoding errors.
+    /// Something went wrong
     Failed,
 
     /// The call failed because of a temporary lack of resources. This could be space resources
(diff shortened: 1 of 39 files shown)


ariel-miculas:
Relates to #221

ariel-miculas:
Maybe I should add a conversion from the enum values to the error strings in https://github.com/capnproto/capnproto-rust/blob/6e421db3b13ead5e92cce68cff186a96492b7d1d/capnp-rpc/src/rpc.rs#L368C3-L368C3 since this test is failing https://github.com/capnproto/capnproto-rust/blob/6e421db3b13ead5e92cce68cff186a96492b7d1d/capnp-rpc/test/test.rs#L339

dwrensha:
> Maybe I should add a conversion from the enum values to the error strings in https://github.com/capnproto/capnproto-rust/blob/6e421db3b13ead5e92cce68cff186a96492b7d1d/capnp-rpc/src/rpc.rs#L368C3-L368C3 since this test is failing
> 
> https://github.com/capnproto/capnproto-rust/blob/6e421db3b13ead5e92cce68cff186a96492b7d1d/capnp-rpc/test/test.rs#L339

I think there should be a conversion from enum values to more verbose messages, and that it should live in an `impl Display`.

capnp/src/lib.rs
```diff
@@ -257,21 +473,23 @@ impl core::convert::From<alloc::str::Utf8Error> for Error {
...
     }
 }
 
 impl core::fmt::Display for Error {
     fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::result::Result<(), core::fmt::Error> {
-        write!(fmt, "{:?}: {}", self.kind, self.description)
+        #[cfg(feature = "alloc")]
+        let result = write!(fmt, "{:?}: {}", self.kind, self.description);
+        #[cfg(not(feature = "alloc"))]
+        let result = write!(fmt, "{:?}", self.kind);
```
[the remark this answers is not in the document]
dwrensha: (Note that you write these messages just fine in an `impl Display` without needing `alloc`.)

capnp/src/schema_capnp.rs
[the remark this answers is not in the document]
dwrensha: I opened https://github.com/capnproto/capnproto-rust/pull/418 to avoid this kind of problem in the future.

capnp/src/lib.rs
```diff
@@ -169,16 +170,210 @@ pub struct Error {
...
 
     /// Human-readable failure description.
+    #[cfg(feature = "alloc")]
     pub description: String,
 }
 
 /// The general nature of an error. The purpose of this enum is not to describe the error itself,
 /// but rather to describe how the client might want to respond to the error.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ErrorKind {
```
dwrensha, line 180: I think that we should add [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) on this, so that we can later add new variants as a non-breaking change. (See [`std::io::ErrorKind`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html) for comparison.)

capnp/src/lib.rs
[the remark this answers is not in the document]
ariel-miculas: I needed a `#[cfg(feature = "alloc")]` guard because the field `description` is under a `#[cfg(feature = "alloc")]` guard. One way to do this would be to get rid entirely of the `description` String and using the enum error variants everywhere (in `capnp-rpc`, `capnp-futures` etc.). I opted to replace the String with enum only in `capnp` to avoid making too many changes. What do you think, should this be done everywhere?

capnp/src/text.rs
```diff
@@ -44,7 +44,7 @@ pub type Reader<'a> = &'a str;
 pub fn new_reader(v: &[u8]) -> Result<Reader<'_>> {
     match str::from_utf8(v) {
         Ok(v) => Ok(v),
-        Err(e) => Err(Error::failed(format!("Text contains non-utf8 data: {e:?}"))),
+        Err(_e) => Err(Error::from_kind(ErrorKind::TextContainsNonUtf8Data)),
```
dwrensha, line 47: Can we make the `ErrorKind` variant carry along the `core::str::Utf8Error` here?

capnp/src/schema.rs
```diff
@@ -70,10 +70,7 @@ impl StructSchema {
         if let Some(field) = self.find_field_by_name(name)? {
             Ok(field)
         } else {
-            Err(crate::Error::failed(format!(
-                "field \"{}\" not found",
```
dwrensha, line 74: It's sad that we lose the field name here.

This makes me think that we should:
1. Rename the `description` field to `extra` (meaning "extra context about error").
2. Add a `Error::extra(&mut self, message: String)` method for populating it. When `alloc` is not enabled, this will just drop `message`.

ariel-miculas: the problem here is that `String` is not available in `no alloc` scenarios, so I need to guard the function itself with `#[cfg(feature = "alloc")]`, and also every call site; but if here is the only case where it's needed, it shouldn't be too bad

ariel-miculas:
I think we can remove "except for error message" from the [documentation](https://docs.rs/capnp/latest/capnp/serialize/fn.read_message_from_flat_slice_no_alloc.html) with this PR. Do you have additional technical comments related to the code?

dwrensha:
> I might open other PRs with #[cfg(feature = "alloc")] in places where allocations happen, it's proven to be quite tricky until now.

Please feel free to add those changes to this PR, if it makes sense to do so.

ariel-miculas:
I need some help with the remaining `cargo clippy` warnings:
```
❯ cargo clippy
warning: unsafe function's docs miss `# Safety` section
  --> capnp/src/private/arena.rs:43:5
   |
43 | /     unsafe fn check_offset(
44 | |         &self,
45 | |         segment_id: u32,
46 | |         start: *const u8,
47 | |         offset_in_words: i32,
48 | |     ) -> Result<*const u8>;
   | |___________________________^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc
   = note: `#[warn(clippy::missing_safety_doc)]` on by default

warning: unsafe function's docs miss `# Safety` section
   --> capnp/src/private/layout.rs:162:5
    |
162 |     pub unsafe fn target(ptr: *const Self) -> *const u8 {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc

warning: unsafe function's docs miss `# Safety` section
    --> capnp/src/private/layout.rs:2918:5
     |
2918 |     pub unsafe fn get_root_unchecked<'b>(location: *const u8) -> PointerReader<'b> {
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc

warning: unsafe function's docs miss `# Safety` section
    --> capnp/src/private/layout.rs:3959:5
     |
3959 | /     pub unsafe fn is_canonical(
3960 | |         &self,
3961 | |         read_head: &Cell<*const u8>,
3962 | |         reff: *const WirePointer,
3963 | |     ) -> Result<bool> {
     | |_____________________^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc

warning: `capnp` (lib) generated 4 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
```

dwrensha:
Do those clippy warnings also get reported on the `master` branch? If so, then cleaning them up can be considered be out-of-scope for this pull request.

ariel-miculas:
> Do those clippy warnings also get reported on the `master` branch? If so, then cleaning them up can be considered be out-of-scope for this pull request.

Ok, I will open a subsequent PR for the clippy changes.

ariel-miculas:
There are some formatting changes with the latest nightly version of rust, I included these changes as a separate commit, let me know if you would like to see them squashed in another commit, I'm not sure what the best practice is in this situation.

dwrensha:
Rebased and merged in 13ccb485e2131fc89ad9d12f1d8ee453502d3f58 087a0fe95a4d4deafa041f561d621beb6b06c793 c410b05220cd04307cc7f3c34669aa1d08b5b446.

I'm going to aim for releasing v0.18 this week.

ariel-miculas:
Kudos for this, I know it was very tedious work

ariel-miculas:
Btw, here's my perl script I've used to (semi)automate the replacements (in case someone wants to do the same for the other error strings): https://gist.github.com/ariel-miculas/e114394eccda9563434c1ad5af19c797

dwrensha:
> I'm going to aim for releasing v0.18 this week.

I'm now thinking that the release will be at least a few weeks from now.

PR #421, opened 2023-07-03
work in progress on supporting setting primitive lists from native Rust slices

dwrensha:


dwrensha:
Closing, because https://github.com/capnproto/capnproto-rust/commit/f4ae39dce25967e919b1bc6f814f8ec9f7942957 eliminates the need for this.

PR #423, opened 2023-07-05
Add NoAllocBufferSegments, suitable for no_alloc environments

ariel-miculas:
It is similar to BufferSegments, but it avoids any allocations.

### DIFF
--- a/capnp/src/serialize.rs
+++ b/capnp/src/serialize.rs
@@ -24,7 +24,7 @@
 //! where each message is preceded by a segment table indicating the size of its segments.
 
 mod no_alloc_slice_segments;
-pub use no_alloc_slice_segments::NoAllocSliceSegments;
+pub use no_alloc_slice_segments::{NoAllocBufferSegments, NoAllocSliceSegments};
 
 #[cfg(feature = "alloc")]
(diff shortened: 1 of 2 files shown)


ariel-miculas:
With this PR I can use capnp in the linux kernel, I've tested it and it works (with the honorable mention of removing the f32/f64 match arms in `stringify.rs`). See https://github.com/ariel-miculas/linux/tree/puzzlefs_rfc for details.

dwrensha:
Thanks!

dwrensha:
I made `NoAllocSliceSegments` a type alias for `NoAllocBufferSegments<&[u8]> in https://github.com/capnproto/capnproto-rust/commit/4015fa0b44423b867cf632e84408078615aba3eb .

dwrensha:
@ariel-miculas in #475 I change the `Deref<Target=[u8]>` to `AsRef<[u8]>`. Please let me know if you foresee any problems with that. I don't expect any problems, because if a type implements the former then it should be easy to add an impl for the latter (if it does not yet already exist).

ariel-miculas:
Looks straightforward to me, thanks for the heads-up.

PR #428, opened 2023-08-21
Support for embedded-io

balderk:
Added support for embedded-io.

Implemented Write, Read, and BufRead for objects with implementations in embedded-io.
This makes capnp easier to use on embedded devices.

### DIFF
--- a/capnp/Cargo.toml
+++ b/capnp/Cargo.toml
@@ -36,8 +36,6 @@ unaligned = []
 # with the Rust standard library.
 std = ["embedded-io?/std"]
 
-
-embedded-io = []
 # If enabled, ReadLimiter will use `AtomicUsize` instead of `Cell<usize>`, allowing
 # message readers to be `Sync`. Note that AtomicUsize is not supported by all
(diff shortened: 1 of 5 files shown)


capnp/Cargo.toml
```diff
@@ -17,11 +17,13 @@ keywords = ["encoding", "protocol", "serialization"]
...
 quickcheck = { version = "1", optional = true }
 
+embedded-io = { version = "0.5.0", default-features = false, optional = true }
+
 [dev-dependencies]
 quickcheck = "1"
 
 [features]
-alloc = []
+alloc = ["embedded-io?/alloc"]
```
dwrensha, line 26: What does this `?` mean?

balderk: If the embedded-io feature is enabled, also enable the embedded-io/alloc feature. 

Without the "?", the embedded-io/alloc feature would be enabled if the "alloc" feature was enabled, even when the embedded-io feature was not.

dwrensha:
Looks good to me.

PR #435, opened 2023-09-13
add AsRef to ease upgrading to 0.18

pcrumley:
Hi! 

I help maintain an open source repo that uses capnp and I'm trying to upgrade it to the latest version v0.18.1. I saw that you added LazyUtf8 support and now everything needs to be a `Reader<T>`: https://dwrensha.github.io/capnproto-rust/2023/09/04/0.18-release.html

I think it is more easy to upgrade & also fits more with the rust ecosystem to add a `From <AsRef<str>>` implementation to the text reader.

Note: This is my first ever commit to this repo and it's possible I missed something or there might be more places where adding this From might make sense.  I also understand that the lack of this `From` implementation may have been a conscious choice to make debugging compilation errors easier, or there may be other things I have overlooked.

### DIFF
--- a/capnp/src/text.rs
+++ b/capnp/src/text.rs
@@ -21,6 +21,7 @@
 
 //! UTF-8 encoded text.
 
+use core::convert::AsRef;
 use core::str;
 
 use crate::Result;
@@ -66,6 +67,16 @@ impl<'a> core::fmt::Debug for Reader<'a> {
     }
 }
 
+impl<'a, T> From<&'a T> for Reader<'a>
+where
+    T: AsRef<str>,
+{
+    #[inline]
+    fn from(value: &'a T) -> Self {
+        Self(value.as_ref().as_bytes())
+    }
+}
+
 impl<'a> From<&'a str> for Reader<'a> {
     #[inline]
     fn from(value: &'a str) -> Self {
@@ -77,12 +88,6 @@ impl<'a> From<&'a [u8]> for Reader<'a> {
     #[inline]
     fn from(value: &'a [u8]) -> Self {
         Self(value)
-    }
-}
-
-impl<'a, const N: usize> From<&'a [u8; N]> for Reader<'a> {
-    fn from(value: &'a [u8; N]) -> Self {
-        Self(&value[..])
     }
 }
 


pcrumley:
I think this actually doesn't work because I had to remove support for `&[u8, N]` because of:
"upstream crates may add a new impl of trait `std::convert::AsRef<str>` for type `[u8; _]` in future versions" 

This will never be added though because implementing this would be unsafe.  I'll leave this PR open but I'm not sure it can be taken and maybe you should just close it.

dwrensha:
Thanks!

I don't think that `impl From<&'a [u8; N]> for text::Reader` is super important. I removed the usages of it just now in https://github.com/capnproto/capnproto-rust/commit/b25b2f0a0841eb46f1cf9838d9707af8168afe26.
Probably we can delete that impl with nobody caring. (Technically that would require a bump to 0.19, but maybe we can claim that its inclusion was a bug?)

Can you tell me more about in what situations you are finding that your new impl would help? I guess that you have some type that implements `AsRef<str>`, and to get that to work with the new capnp interface, you currently need to add an `into()` *and* a type annotation, whereas if we had the new impl, you'd be able to turn an `as_ref()` into an `into()`?

pcrumley:
So I opened this PR because we had a capnp that had `List(Text)` as one of its items, and in our Rust codebase we fill it out by having a `Vec<String>` which we iterate over to fill that list up. In the current version we can either do
```rust
let my_values = vec![ "Foo".to_string(), "bar".to_string()];
for (i, v) in my_values.iter().enumerate() {
    my_list.set(i as u32, v.as_str().into())
}
```
whereas after this PR we simply would have to do
```rust
let my_values = vec![ "Foo".to_string(), "bar".to_string()];
for (i, v) in my_values.iter().enumerate() {
    my_list.set(i as u32, v.into())
}
```

dwrensha:
Does this work? (It's a little bit more concise.)
```rust
let my_values = vec![ "Foo".to_string(), "bar".to_string()];
for (i, v) in my_values.iter().enumerate() {
    my_list.set(i as u32, v[..].into())
}
```

pcrumley:
The more i think about this, I think it might make more sense to add an `AsRef<[u8]>` instead of `AsRef<Str>`, in which case i think removing the `impl From<&'a [u8; N]> for text::Reader` would not be a breaking change as it implements `AsRef<[u8]>`. It also seems to go more along the spirit of the change of the lazyUtf8 validation.

dwrensha:
We should also at least consider using `Deref` instead of `AsRef`, like
```rust
impl<'a, T> From<&'a T> for Reader<'a>
where
    T: std::ops::Deref<Target=str>
{
    #[inline]
    fn from(value: &'a T) -> Self {
        Self(value[..].as_bytes())
    }
}
```
or
```rust
impl<'a, T> From<&'a T> for Reader<'a>
where
    T: std::ops::Deref<Target=[u8]>
{
    #[inline]
    fn from(value: &'a T) -> Self {
        Self(&value[..])
    }
}
```

pcrumley:
So i started with `Deref<Target=str>` locally, and it breaks the build of `capnp`. I also know people have feelings about deref coercion so I thought I should start more modestly.

dwrensha:
> but unlike AsRef<Str>, i don't believe it would handle Cow<_, str>

That sounds like a pretty strong argument for `AsRef<str>`. (I've never used `Cow`, but I get the sense that it can be pretty useful sometimes.)

pcrumley:
yeah you only get one bite at this apple as the traits are conflicting. My feeling is if you had started with LazyUtf8 it would make more sense to just use `AsRef<[u8]>`, as that makes the most sense to me. Because you are upgrading from when it used to take a `str`, it might make more sense to use `AsRef<str>` because that will kind of keep an easier path from before. But then if you do in fact want to enable sending invalid utf8 in the Text field it will be more burdensome.

Depends on the design goal of the lib.

dwrensha:
> But then if you do in fact want to enable sending invalid utf8 in the Text field it will be more burdensome.

We want it to be possible to *read* invalid text, but we don't particularly want it to be convenient to *write* invalid text.

-----

I was at first eager to add one of these new impls, because I thought that it would make stuff like this work:

```rust
for i in 0.. my_list.len() {
    my_list.set(i, format!("element {i}").into())
}
```

but I tried it, and it turns out that still does not work. We would need to do

```rust
for i in 0.. my_list.len() {
    my_list.set(i, (&format!("element {i}")).into())
}
```

which I think is no better than the current status quo of 

```rust
for i in 0.. my_list.len() {
    my_list.set(i, format!("element {i}")[..].into())
}
```
So adding `impl where T : AsRef<str>` has less payoff than I had originally thought.

I'm inclined to:
 1. deprecate the `impl<'a, const N: usize> From<&'a [u8; N]> `
 2. on the 0.19 version bump (not urgently), add the `impl<'a, T> From<&'a T> for Reader<'a> where T: AsRef<str>`

pcrumley:
@dwrensha If that is something you are interested in supporting, I think it is possible by holding a `Cow` inside of the reader.

If you would be interested in an implementation, I could take a crack at it. I would recommend giving it the old college try before you add one of these implementations.

pcrumley:
Supporting something like that workflow should be possible with `Cow`, but then that would require std which i think you have made a choice not to support.

Just to circle back I was able to get  deref working with
```rust
impl<'a, T> From<&'a T> for Reader<'a>
where
    T: Deref<Target = str>,
{
    #[inline]
    fn from(value: &'a T) -> Self {
        Self(value[..].as_bytes())
    }
}
```

which is what I would recommend over AsRef i suppose. It doesn't solve your issue though.

dwrensha:
Another problem of implementing `text::Reader` in terms of `Cow` is that it would presumably not allow for pattern matching, like: https://github.com/capnproto/capnproto-rust/blob/20741d11f543d672f3a88d5633652f4af651b0ec/capnpc/test/test.rs#L1118-L1120

dwrensha:
#472 gives text field setter methods a type parameter `<T : SetPointerBuilder<text::Owned>>`, which makes it possible for the input value to be either a `text::Reader`, or a `&str` or a `&String`. I expect that this should make this PR unneeded. (But it comes at a cost of requiring some more churn in user code.)

Issue #445, opened 2023-11-08
`capnp_futures::serialize_packed::try_read_message()` is blocked forever with 100% cpu usage when the input stream is closed


paulnice:
When I connect to a server with a tcp socket, read from it with `capnp_futures::serialize_packed::try_read_message()`, and server disconnect my socket; my client is blocked on the `try_read_message()` forever and using 100% of cpu.

PR #446, opened 2023-11-08
Catch EOF properly in `serialize_packed` AsyncRead

paulnice:
Problem: `capnp_futures::serialize_packed::try_read_message()` blocks forever on an input stream which is completed and reports EOF to us. The library uses 100% cpu while is doing that.

Solution: As soon as the library reads 0 bytes from the stream, interpret it as EOF and return `Poll::Ready(Ok(0))` as expected by AsyncRead trait.

Issue: https://github.com/capnproto/capnproto-rust/issues/445

### DIFF
--- a/capnp-futures/src/serialize_packed.rs
+++ b/capnp-futures/src/serialize_packed.rs
@@ -100,6 +100,9 @@ where
                     match Pin::new(&mut *inner).poll_read(cx, &mut buf[*buf_pos..2])? {
                         Poll::Pending => return Poll::Pending,
                         Poll::Ready(n) => {
+                            if n == 0 {
+                                return Poll::Ready(Ok(0));
+                            }
                             *buf_pos += n;
                             if *buf_pos >= 2 {
                                 let tag = buf[0];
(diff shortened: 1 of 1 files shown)


dwrensha:
Thanks!

paulnice:
Thank you, @dwrensha . Is it possible to backport these changes to other versions? Like 0.14?

dwrensha:
Added a test in https://github.com/capnproto/capnproto-rust/commit/45744ca2899d824bc2d3d98c1cdbc5f6731f61dc.

dwrensha:
> Is it possible to backport these changes to other versions? Like 0.14?

It's possible, but takes some effort. What's keeping you from upgrading?

paulnice:
> > Is it possible to backport these changes to other versions? Like 0.14?
> 
> It's possible, but takes some effort. What's keeping you from upgrading?

Big codebase, many crates, which depends on capnp, several teams. We just recently moved to 0.14 and that was long and painful.

paulnice:
@dwrensha I can work on backporting my fix to 0.14 version. I can create a PR to try. Will that work for you?

dwrensha:
There is at least one other bug that needs to be fixed: #447

paulnice:
@dwrensha I can work on it, if you'd like. I think capnp-futures requires to have more comprehensive tests

PR #452, opened 2023-11-17
Introduce InnerBufferAllocator

kolegs:
Hi,

I've added `InnerBufferAllocator` to simplify some cases. Just sharing in case we want to also have it in our repository.

Might be that I am not aware of something, but it simplifies the things when you want to return `TypedReader` from the function using the buffer created on stack. Using `SingleSegmenetAllocator` is much complicated in that case, because you need to somehow make sure that the buffer survives.

@dwrensha comments are welcomed and I can also write some unit tests if needed. Feel free to modify it if needed

### DIFF
--- a/capnp/src/message.rs
+++ b/capnp/src/message.rs
@@ -881,7 +881,7 @@ unsafe impl<'a> Allocator for ScratchSpaceHeapAllocator<'a> {
 /// part of the buffer upon `deallocate_segment()`.
 ///
 /// You can reuse a `SingleSegmentAllocator` by calling `message::Builder::into_allocator()`,
-/// or by initally passing it to `message::Builder::new()` as a `&mut SingleSegmentAllocator`.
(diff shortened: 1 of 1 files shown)


dwrensha:
I think what we really want is to generalize `SingleSegmentAllocator` to
```rust
pub struct SingleSegmentAllocator<T> where T : DerefMut<Target=[u8]> {
    segment: T,
    segment_allocated: bool,
}
```
That's analogous to what we currently do for no-alloc reading: https://github.com/capnproto/capnproto-rust/blob/9b2e8fab9f50cc6e5ef61f23a2b75680cfa0f36a/capnp/src/serialize/no_alloc_buffer_segments.rs#L105-L108

I'm going to see if this works, as part of the v0.19.0 release...

dwrensha:
I opened #474. It seems that `AsMut` makes more sense here than `DerefMut`.

dwrensha:
Closing, because #474 has landed. You can now use `SingleSegmentAllocator<[u8; N]>`.

PR #456, opened 2023-11-27
Implement this_client()

### LINKED ISSUE #87: [RPC] add way to get a self-reference from within a method call
capnproto-c++ has [`thisCap`](https://github.com/capnproto/capnproto/blob/44f5f5686962f3c8ff95618785d1998efd63a779/c%2B%2B/src/capnp/capability.h#L335).

I imagine we could achieve something similar by adding a `get_self() -> Client` method to [`ParamsHook`](https://github.com/capnproto/capnproto-rust/blob/dd67e97405bcd23082f221f7a1a2d6b57de1e47d/capnp/src/private/capability.rs#L102) and [`Params`](https://github.com/capnproto/capnproto-rust/blob/dd67e97405bcd23082f221f7a1a2d6b57de1e47d/capnp/src/capability.rs#L155).

ranfdev:
fixes #87.

One question though: this currently works well for a struct implementing a single interface. 
If a single struct implements multiple interfaces, there will be multiple `this_client` implementations, one for each interface.
I'm not sure how this is going to handle multiple interfaces...

For example:
```rust
struct Calculator{}
impl calculator_add::Server for Calculator {
  fn add(&mut self, params, results) {...} 
}
impl calculator_sub::Server for Calculator {
  fn sub(&mut self, params, results) {
    let client = calculator_sub::Server::this_client(self); // Works perfectly.
    let client = calculator_add::Server::this_client(self); // OUCH, we are trying to get a `calculator_add::Client` from a `calculator_sub::Client`.
  } 
}
```

If the two interfaces share the same `ClientHook` anyway, this is fine. Do they? Else, it's going to be a problem.

### DIFF
--- a/capnp-rpc/src/local.rs
+++ b/capnp-rpc/src/local.rs
@@ -368,15 +368,38 @@ where
         let inner = self.inner.clone();
         Promise::from_future(async move {
             let f = {
+                let this_hook = Self { inner };
+
+                // We need to provide a way to get the corresponding ClientHook of a Server.
+                // Passing the ClientHook inside `Params` would require a lot of changes, breaking compatiblity
+                // with existing code and ruining the developer experience, because `Params` would end up containing more generics.
+                // Instead, I'm passing the `ClientHook` through a global static variable, CURRENT_THIS.
+                // This operation will be called for every method call, so we cannot allocate a `Box<dyn ClientHook>` everytime.
+                // To save us from allocating a `Box` even when we don't access CURRENT_THIS, we set the static variable to a closure
+                // returning `Box<dyn ClientHook>`, making the allocation "lazy".
+                let prev = unsafe {
+                    // This is a gimmick to make Rust happy. We can only set static values on a `static` variable.
+                    // `static_hook` is not actually static, because it's referencing a variable on the stack.
+                    // Before `this_hook` is dropped, we remove `static_hook` from CURRENT_THIS.
+                    let static_hook = &this_hook as *const (dyn ClientHook + 'static);
+                    capnp::private::capability::CURRENT_THIS.replace(Some(&{
+                        move || (&*static_hook as &dyn ClientHook).add_ref()
+                    }
+                        as *const (dyn Fn() -> Box<(dyn ClientHook + 'static)> + 'static)))
+                };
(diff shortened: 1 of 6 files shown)


capnp-rpc/src/local.rs
```diff
@@ -368,15 +368,38 @@ where
         let inner = self.inner.clone();
         Promise::from_future(async move {
             let f = {
+                let this_hook = Self { inner };
+
+                // We need to provide a way to get the corresponding ClientHook of a Server.
+                // Passing the ClientHook inside `Params` would require a lot of changes, breaking compatiblity
+                // with existing code and ruining the developer experience, because `Params` would end up containing more generics.
```
dwrensha, line 375: What if `this_client()` returned an untyped capability -- perhaps just a `Box<dyn ClientHook>`? That would not require any more generics, would it?
(It looks to me like that's how capnproto-c++ implements `thisCap()`.)

ranfdev: Having a `params.this_client()` returning `Box<dyn ClientHook>` would be possible but it has the following drawbacks:
- the user needs to manually cast the client. 
- `params` is passed by value, and can be moved around and stored eventually for a `'static` lifetime. That means `params` would need to own a copy of the current client, requiring a `Box` allocation for each method call. Unless we store a `Weak` reference to the `ClientHook`, but then `params.this_client()` needs to return an `Option<T>`, because the `params` we own may reference a dead `ClientHook`.

Returning an `Option<Box<dyn ClientHook>>` isn't great compared to a fully typed `calculator::op_add::Client`

ranfdev: An alternative API which doesn't require unsafe nor a static variable is
`params.client_for(self)`. (Where self is actually `&mut self`, because we are inside a method call).

By requiring `&mut self` as a parameter, we restrict the ability to get a client while we are  inside the method call, and we also have the type information required to return a typed Client. 

It's fully safe and typed Rust.

dwrensha:
I implemented an alternative in #473.

dwrensha:
https://github.com/capnproto/capnproto-rust/pull/595 included a simpler solution: methods get `self: Rc<Self>`, which can be passed to `capnp_rpc::new_client_from_rc()`.

PR #458, opened 2023-11-29
add clippy linting CI target

danieleades:
adds a clippy linting CI target.

Couple of questions on this-

- what is the best possible combination of features to use for linting?
- would you prefer to have the compiler version used for clippy linting be pinned? (that prevents build failures on PRs due to new lints, but means the version must be manually updated at intervals)


danieleades:
@dwrensha this is complete and ready for review.

Note that to get the linting to pass i've had to temporarily suppress a type complexity lint.
I'm not in a position to create the type aliases to make this lint happy, as the types are impenetrable to me (which illustrates the value of the lint!)
I leave that as an exercise ;)

capnp-rpc/src/rpc.rs
```diff
@@ -1031,61 +1031,65 @@ impl<VatId> ConnectionState<VatId> {
...
+                                    );
+                                    question_ref.borrow_mut().fulfill(Promise::ok(response));
+                                }
+                                return_::Exception(e) => {
+                                    let tmp =
+                                        question_ref.upgrade().expect("dangling question ref?");
+                                    tmp.borrow_mut().reject(remote_exception_to_error(e?));
+                                }
+                                return_::Canceled(_) => {
+                                    Self::send_unimplemented(&connection_state, message.as_ref())?;
```
danieleades, line 1057: i've used 'as_ref' to dereference the box to the inner type.

an alternative would be to `impl IncomingMessage for Box<dyn IncomingMessage>`

capnp/Cargo.toml
```diff
@@ -40,3 +40,10 @@ std = ["embedded-io?/std"]
 # message readers to be `Sync`. Note that AtomicUsize is not supported by all
 # rustc targets.
 sync_reader = []
+
+#[lints]
+#workspace = true
```
dwrensha, line 45: Why is this commented out? I guess the idea is that we intend to delete the `lints.clippy` block below, and when that happens, we'll uncomment this?

.github/workflows/ci.yml
```diff
@@ -137,3 +137,20 @@ jobs:
...
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Cap'n Proto
+        run: |
+            export DEBIAN_FRONTEND=noninteractive
+            sudo apt-get install -y capnproto libcapnp-dev
+
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@nightly
```
dwrensha, line 151: I think pinning to a rustc version would be better. 1.74 is fine.

capnp-rpc/Cargo.toml
```diff
@@ -21,3 +21,9 @@ features = ["std"]
 [dependencies]
 capnp-futures = { version = "0.18.0", path = "../capnp-futures" }
 capnp = {version = "0.18.0", path = "../capnp"}
+
+#[lints]
+#workspace = true
```
dwrensha, line 26: Why commented out?

danieleades: as above. this should be uncommented as soon as the type complexity lint is addressed.

Would you prefer i remove the commented sections?

Issue #455, opened 2023-11-27
Improve performance of Reader get_named


quartox:
I was doing some performance testing of code that serializes a large number of capnp messages. The biggest bottleneck is `capnp::dynamic_struct::Reader.get_named` because it does an [`O(N)` lookup](https://github.com/capnproto/capnproto-rust/blob/master/capnp/src/schema.rs#L60-L61) of the field names, converts them to text and compares them to the input.

I was wondering if you would be open to a contribution that improves the performance of this method.

Naively I was thinking of solving this with a `HashMap` that populates the names of fields as keys and fields as values. This could be populated at runtime after an `O(N)` lookup.

dwrensha:
Thanks for the report.

We have a TODO in the code about this:
https://github.com/capnproto/capnproto-rust/blob/eb5b768657f53bb4a3e9378e154af24b6f02cfac/capnp/src/introspect.rs#L217

capnproto-c++ has implemented this via binary search:
https://github.com/capnproto/capnproto/blob/537fe97d6ab1962e92e15b8a16f60439a63b90f5/c%2B%2B/src/capnp/raw-schema.h#L184-L186

https://github.com/capnproto/capnproto/blob/537fe97d6ab1962e92e15b8a16f60439a63b90f5/c%2B%2B/src/capnp/schema.c%2B%2B#L458-L481

quartox:
I started [adding a hashmap in a fork](https://github.com/capnproto/capnproto-rust/commit/8041d3d516654c7f42e72de3a21e5c46ae9f94a8) but I cannot re-generate the code because the previously generated code does not compile with the new field:

```
error[E0063]: missing field `members_by_name` in initializer of `RawStructSchema`
     --> capnp/src/schema_capnp.rs:14437:21
      |
14437 |                     crate::introspect::RawStructSchema {
      |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `members_by_name`
```

I tried removing the old generated file but that led to other errors for files that depend on that file. Any advice on how to add a new field and re-generate the generated files?

I also went down the path of a hashmap since the C++ code also had a todo suggesting this. We lose the Copy trait on RawStructSchema and I do not know the implications of that. This seems like a viable path but willing to redo doing binary search instead.

dwrensha:
> I tried removing the old generated file but that led to other errors for files that depend on that file. Any advice on how to add a new field and re-generate the generated files?

[...]

PR #469, opened 2024-01-04
Use binary search for find capnp struct field by name

quartox:
Following discussion in https://github.com/capnproto/capnproto-rust/issues/455 I added `members_by_name` to `RawStructSchema` which contains a sorted list of tuples with the name of the field and the index of the field. This is used for binary search when finding the field in a capnp schema using the name of the field copying the [C++ binary search algorithm](https://github.com/capnproto/capnproto/blob/537fe97d6ab1962e92e15b8a16f60439a63b90f5/c%2B%2B/src/capnp/schema.c%2B%2B#L458-L481).

I need to test the performance improvement of this change.

### DIFF
--- a/capnp/src/introspect.rs
+++ b/capnp/src/introspect.rs
@@ -203,7 +203,7 @@ primitive_introspect!(f64, Float64);
 
 /// Type information that gets included in the generated code for every
 /// user-defined Cap'n Proto struct.
-#[derive(Clone)]
+#[derive(Copy, Clone)]
 pub struct RawStructSchema {
     /// The Node (as defined in schema.capnp), as a single segment message.
     pub encoded_node: &'static [crate::Word],
@@ -214,8 +214,8 @@ pub struct RawStructSchema {
     /// Map from discriminant value to field index.
     pub members_by_discriminant: &'static [u16],
 
-    // Map field name to field index.
-    pub members_by_name: std::collections::HashMap<&'static str, u16>,
+    // Sorted array of field names with matching field index.
+    pub members_by_name: &'static [(&'static str, u16)],
 }
 
 /// A RawStructSchema with branding information, i.e. resolution of type parameters.
--- a/capnp/src/introspect.rs
+++ b/capnp/src/introspect.rs
@@ -215,7 +215,7 @@ pub struct RawStructSchema {
     pub members_by_discriminant: &'static [u16],
 
     // Sorted array of field names with matching field index.
-    pub members_by_name: &'static [(&'static str, u16)],
+    pub members_by_name: &'static [u16],
 }
 
 /// A RawStructSchema with branding information, i.e. resolution of type parameters.
--- a/capnp/src/introspect.rs
+++ b/capnp/src/introspect.rs
@@ -203,7 +203,7 @@ primitive_introspect!(f64, Float64);
 
 /// Type information that gets included in the generated code for every
 /// user-defined Cap'n Proto struct.
-#[derive(Copy, Clone)]
+#[derive(Clone)]
 pub struct RawStructSchema {
     /// The Node (as defined in schema.capnp), as a single segment message.
     pub encoded_node: &'static [crate::Word],
@@ -213,8 +213,9 @@ pub struct RawStructSchema {
 
     /// Map from discriminant value to field index.
     pub members_by_discriminant: &'static [u16],
-    //
-    // TODO: members_by_name, allowing fast field lookup by name.
+
+    // Map field name to field index.
+    pub members_by_name: std::collections::HashMap<&'static str, u16>,
 }
 
 /// A RawStructSchema with branding information, i.e. resolution of type parameters.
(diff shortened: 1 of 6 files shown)


quartox:
Performance testing on this commit cuts time spent in `find_field_by_name` from 15% to 3.5%. I don't trust the exact numbers but clearly a significant performance improvement. Annoyed that I can't figure out how to do this O(1) but this is a big help.

capnp/src/introspect.rs
```diff
@@ -213,8 +213,9 @@ pub struct RawStructSchema {
 
     /// Map from discriminant value to field index.
     pub members_by_discriminant: &'static [u16],
-    //
-    // TODO: members_by_name, allowing fast field lookup by name.
+
+    // Sorted array of field names with matching field index.
+    pub members_by_name: &'static [(&'static str, u16)],
```
dwrensha, line 218: The direct translation of capnproto-c++ would make this a `&[u16]`: https://github.com/capnproto/capnproto/blob/537fe97d6ab1962e92e15b8a16f60439a63b90f5/c%2B%2B/src/capnp/raw-schema.h#L184

Is there a reason that you did it this way? The `&str` values that you are storing here should be cheaply retrievable via `StructSchema::get_fields()`.

quartox: Did not know it was cheap to get that, thanks for the tip. I saw some calls to `Reader::get_text` in the benchmarking flamegraph that seemed expensive, but I think that was only because of the total number of calls to that function (which we already reduced).

Only problem now is that we don't have `PartialOrd` to compare the field name passed by the user and `text::Reader`.

quartox: I tested on [this branch](https://github.com/capnproto/capnproto-rust/compare/master...vectra-ai-research:capnproto-rust:get-name) and using `StructSchema::get_fields` made it slower than the O(N) implementation. Most of the time is spent getting the `FieldList` followed by `get_name`.

dwrensha: In that branch, you are converting to `utf8` before making a comparison:
```rust
} else if candidate_name.to_str()? < name {
```
If you directly make the comparison on `capnp::text::Reader`, how much faster does that make things? (Implementing `PartialOrd` for `text::Reader` should be easy, because `&[u8]` already implements `PartialOrd`.)

It might also help to add `#[inline]` attributes on various methods, like `StructSchema::get_proto()`. (I don't particularly expect this to help, unless it somehow enables cross-crate inlining, but it seems worth a try.)

quartox: I added the `PartialOrd` trait to the for the `capnp::text::Reader` but the less than comparison with `&str` does not compile. The compiler complains that they are different types. I was surprised since the equal to comparison compiles. For what it is worth the `to_str` call adds a negligible amount to the flamegraph compared to `get_name`. I obviously am seeking performance improvement but ideally not adding any unnecessary complexity that makes it more confusing to read.

I will try inlining `StructSchema::get_proto`. I was also considering an `Option` field in the `StructSchema` to cache the `FieldList` after the first call to `StructSchema::get_fields`.

dwrensha: > but the less than comparison with &str does not compile... I was surprised since the equal to comparison compiles

Here are the impls for `PartialEq` which make that possible:

https://github.com/capnproto/capnproto-rust/blob/e476eaaeeee8ef3c079782d7893547c8e9a2aee0/capnp/src/text.rs#L48-L58

I think something similar can be done with `PartialOrd`.

dwrensha: added in f6adc78bcf892f4bab9bea5cecf2e6bb6a3627ac

quartox: Neither change improved performance. I have attached a screenshot of the latest flamegraph. The `ParitalOrd` takes as long as the `to_str` comparison.

Screenshot 2024-01-08 at 9 55 50 AM

dwrensha: Thanks for the flame graph.

To summarize:
 1. Status quo is that `get_field_by_name` accounts for 43% of your running time.
 2. Implementing binary search on `members_by_name: &'static [u16]` brings that down to 29%.
 3. Implementing binary search on `members_by_name: &'static [(&'static str, u16)]` brings it down to 19%.
 
The cost of (3) is that it increases the size of the generated code, because then fields names get included twice (once in the serialized schema, and once in this new array). Moreover, I'm resistant to diverging from the capnproto-c++ way of doing things.

---

Other `#[inlines]` could potentially help, like maybe `FieldList::get()` and `struct_list::Reader::get()`.

Is there an easy way that you could share a benchmark with me so that I can play around with it?

dwrensha: @quartox roughly how many fields do your structs have?

quartox: I was going to try to share the benchmark but it isn't as simple as I had hoped. I was going to construct a schema with similar properties and generate fake data. I haven't had much success using the command line capnp json to binary converter with as much nesting as is happening in my real schemas. I will see if I can find another way to do this.

I will clean up this PR to use `members_by_name: &'static [u16]` and then we can open another issue to discuss the other performance improvements. Ultimately speeding up the other portions of the code would be more beneficial.

I may explore a static hashmap since the C++ implementation has a todo to make that improvement, but if you want to match what they currently do then we may not adopt it.

dwrensha: I'm considering making a version-bump release soon, including #472 as well.

I like the `members_by_name: &'static [u16]` solution. Can you submit a pull request for that?

dwrensha: Ah, I should have noticed that! That should be sufficient. Thanks!

PR #473, opened 2024-01-12
implement this_cap() as a method of Params

### LINKED ISSUE #87: [RPC] add way to get a self-reference from within a method call

dwrensha:
Fixes #87.

### DIFF
--- a/capnp-rpc/src/local.rs
+++ b/capnp-rpc/src/local.rs
@@ -62,14 +62,20 @@ impl ResponseHook for Response {
 struct Params {
     request: message::Builder<message::HeapAllocator>,
     cap_table: Vec<Option<Box<dyn ClientHook>>>,
+    this_cap: Box<dyn ClientHook>,
 }
 
 impl Params {
     fn new(
         request: message::Builder<message::HeapAllocator>,
         cap_table: Vec<Option<Box<dyn ClientHook>>>,
+        this_cap: Box<dyn ClientHook>,
     ) -> Self {
-        Self { request, cap_table }
+        Self {
+            request,
+            cap_table,
+            this_cap,
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


dwrensha:
One downside to this approach: it adds an extra heap allocation on the server side of every remote procedure call, because we do `client.add_ref()` to populate the `this_cap` field of `Params`.

The natural way to avoid that allocation would be to somehow pass a `&Box<dyn ClientHook>` instead. That could be accomplished by adding a new `&Box<dyn ClientHook>` parameter to every `Server` method (i.e. in addition to the existing `Params` and `Results` parameters), but doing so would add a lot of boilterplate for end users. A better approach might be to add a new `CallContext<'a>` struct (similar to what capnproto-c++ has) that holds the `Params`, `Results`, and also a `&'a Box<dyn ClientHook>` representing `this_cap`.

dwrensha:
https://github.com/capnproto/capnproto-rust/pull/595 includes a simpler solution: methods get `self: Rc<Self>`, which can be passed to `capnp_rpc::new_client_from_rc()`.

PR #476, opened 2024-01-15
Avoid external deps other than the capnp tool for regeneration of schemas

bjorn3:
On my Debian install `/capnp/c++.capnp` isn't resolved even after installing the capnproto package that includes the capnp tool used for parsing the capnp schemes.


dwrensha:
Thanks!

PR #485, opened 2024-02-20
remove unused imports flagged by latest nightly

dwrensha:


dwrensha:
The failure is with
```
cargo test --no-default-features --features alloc -p capnp
```

PR #492, opened 2024-05-07
docs: remove usage of deprecated tokio_core in docs

SteveLauC:
### What does this PR do

Update the documentation of `RpcSystem`

1. `tokio_core` has been deprecated, so replace `tokio_core::reactor::handle::spawn()` with `tokio::task::spawn_local()`.
2. Explicitly document it that `RpcSystem` is not `Send` (so one needs `spawn_local()` rather than `spawn()`).


dwrensha:
Thanks!

PR #497, opened 2024-05-12
Forbid use of as_slice when unaligned feature is enabled

as-com:
Somewhat related to #496 - if the user enables the `unaligned` feature, undefined behavior may result from use of `as_slice()` if the type is larger than 1 byte.

I added a compile-time check that panics in a const fn for a more useful error message and to keep the API the same. This also adds support for `as_slice()` on big-endian targets so long as the lists' elements are 1 byte or less.

Possible alternatives include:

- Returning `None` if the buffer is unaligned, but still returning a slice if it happens to be aligned
- Always returning `None` if the `unaligned` feature is enabled for lists of multi-byte elements
- `#[cfg(...)]`-gating the entire method

### DIFF
--- a/capnp/src/primitive_list.rs
+++ b/capnp/src/primitive_list.rs
@@ -112,9 +112,14 @@ impl<'a, T: PrimitiveElement> Reader<'a, T> {
         }
     }
 
-    #[cfg(target_endian = "little")]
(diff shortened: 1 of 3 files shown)


dwrensha:
Thanks!

PR #499, opened 2024-05-16
fix bug where get_named() could panic on nonexistent fields

### LINKED ISSUE #498: Dynamic parsing of fields may panic
When I use the has_named interface `some_struct.has_named("some_field_name")`, sometimes it's panic.

```
thread 'main' panicked at XXXX\capnp-0.19.2\src\schema.rs:76:31:
index out of bounds: the len is 4 but the index is 4
```

Then I check the code:
https://github.com/capnproto/capnproto-rust/blob/master/capnp/src/schema.rs
```rust
pub fn find_field_by_name(&self, name: &str) -> Result<Option<Field>> {
        let fields = self.get_fields()?;
        let mut lower: usize = 0;
        let mut upper: usize = self.raw.generic.members_by_name.len();
        let mut mid: usize = (lower + upper) / 2;
        let mut candidate_index = self.raw.generic.members_by_name[mid];
        let mut candidate_name = fields.get(candidate_index).get_proto().get_name()?;

        while lower < upper {
            use core::cmp::Ordering;
            match (&name).partial_cmp(&candidate_name) {
                Some(Ordering::Equal) => return Ok(Some(fields.get(candidate_index))),
                Some(Ordering::Greater) => lower = mid + 1,
                Some(Ordering::Less) => upper = mid,
                None => unreachable!(),
            }
            mid = (lower + upper) / 2;
            candidate_index = self.raw.generic.members_by_name[mid];
            candidate_name = fields.get(candidate_index).get_proto().get_name()?;
        }
        Ok(None)
    }
```

I'm guessing that `mid` might be equal to `upper`, which would cause the array to go out of bounds. I don't know if my findings are correct.

dwrensha:
Fixes #498.

### DIFF
--- a/capnp/src/schema.rs
+++ b/capnp/src/schema.rs
@@ -60,11 +60,12 @@ impl StructSchema {
         let fields = self.get_fields()?;
         let mut lower: usize = 0;
         let mut upper: usize = self.raw.generic.members_by_name.len();
-        let mut mid: usize = (lower + upper) / 2;
-        let mut candidate_index = self.raw.generic.members_by_name[mid];
-        let mut candidate_name = fields.get(candidate_index).get_proto().get_name()?;
 
         while lower < upper {
+            let mid: usize = (lower + upper) / 2;
+            let candidate_index = self.raw.generic.members_by_name[mid];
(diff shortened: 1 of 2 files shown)


dwrensha:
cc @quartox

PR #501, opened 2024-05-27
docs: update 2 old links

SteveLauC:
## What does this PR do

Update 2 old links.

The old links are not dead, users will be redirected to the new links, but why not update them :>


dwrensha:
Thanks!

Issue #495, opened 2024-05-10
capnp_rpc does not work on monoio


SteveLauC:
Hi:)

I am playing the `hello-world` example on [monoio](https://github.com/bytedance/monoio) (A runtime that is based on io_uring), but both server and client sides block, and from my observation, `.await`ing `RpcSystem` won't finish and simply blocks, so I come here to seek some help:>

You can find the code [here](https://github.com/SteveLauC/capnp_rpc_monoio), in README, I documented the [changes that I have made to the original code](https://github.com/SteveLauC/capnp_rpc_monoio?tab=readme-ov-file#changes-made-to-the-original-code) and [some observations that I found while debugging the program](https://github.com/SteveLauC/capnp_rpc_monoio?tab=readme-ov-file#things-i-observed-both-server-and-client-side), hope they will help a bit.

dwrensha:
Interesting. What happens if you try a simpler example that uses only [capnp-futures](https://github.com/capnproto/capnproto-rust/tree/master/capnp-futures) (and no RPC)? Does that hit a similar problem?

SteveLauC:
Hi! Thanks for your help and response!

I wrote a single TCP server that would echo the message received from client (pretty similar to the capnp-rpc hello-world example) using capnp-futures and monoio ([repo](https://github.com/SteveLauC/capnp_futures_monoio)), and it look like it is working pretty well:

```sh
$ cargo b
$ ./target/debug/capnp_futures_monoio server 127.0.0.1:8100
INFO: server listening on 127.0.0.1:8100
```
```sh
$ ./target/debug/capnp_futures_monoio client 127.0.0.1:8100 hello
```
```sh
$ ./target/debug/capnp_futures_monoio server 127.0.0.1:8100
INFO: server listening on 127.0.0.1:8100
INFO: Server received hello
```

SteveLauC:
Well, it turns out that it WORKS on monoio!

It blocks in my repo because I .awaited the `RpcSystem`, which should not be done, so this issue does not exist, closing.

dwrensha:
Ah, right. The `RpcSystem` needs to be driven, but if you `await` it in the main task then it will never finish. You need to spawn a new task for it.

PR #502, opened 2024-05-27
docs: document capnrpc/src/lib.rs

SteveLauC:
## What does this PR do

Document most types/interfaces in `capnp-rpc/src/lib.rs`.

## Why

I started working on #495, the first thing was to read the source code, then I found these are a lot of concepts that I am not aware of, and they do not have documents. Considering this is a Rust port is the C++ implementation, authors of this port should be already familiar with those concepts, this is understandable. 

But I am completely new to them, so I took a look at the C++ repo, found the concept I wanted to understand, and copied the documents into the Rust code.

For stuff that already exists in the C++ impl, the documents are directly pasted, documents of other things are added based on my understanding.


dwrensha:
Thanks!

PR #521, opened 2024-09-23
add dynamic_value::Reader::downcast_struct()

dwrensha:
I wanted to make a blanket impl of the `DowncastReader` trait for all structs, but I don't see a way to get all the trait bounds/associated types to work out. So this is the next-best thing.


dwrensha:
@kvap does this work for you?

kvap:
@dwrensha this works perfectly, thanks!

PR #526, opened 2024-10-19
Allow to retrieve `ReaderSegments` object from reader objects

### LINKED ISSUE #524: Serializing from borrowed TypedReader
Hello,

I happen to face a situation where in some API I have a borrowed TypedReader and I need to serialize it.

I might have missed something, but it seems to serialize from a reader, I need access to the ReaderSegments which I can only obtain by doing typed_reader.into_inner().into_segments() which is consuming the input TypedReader.
However the serializing functions (the _to_words ones) actually accept a borrowed ReaderSegments.

Is there a reason for not exposing a borrowed ReaderSegments?
If we added a method like `get_reader_segments(&self) -> &S` it should fix my issue.

Thank you for your help.

marmeladema:
Closes #524

### DIFF
--- a/capnp/src/private/arena.rs
+++ b/capnp/src/private/arena.rs
@@ -82,6 +82,10 @@ where
     pub fn into_segments(self) -> S {
         self.segments
     }
+
+    pub(crate) fn get_segments(&self) -> &S {
+        &self.segments
+    }
 }
 
 impl<S> ReaderArena for ReaderArenaImpl<S>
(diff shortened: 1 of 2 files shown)


capnp/src/private/arena.rs
```diff
@@ -82,6 +82,10 @@ where
     pub fn into_segments(self) -> S {
         self.segments
     }
+
+    pub(crate) fn get_segments(&self) -> &S {
```
marmeladema, line 86: Not sure if this one should be public or not so i've kept it private for now.

PR #527, opened 2024-11-07
capnp-futures: Use more fine-grained futures-* deps

Gelbpunkt:
The large futures crate pulls in a bunch of sub-crates that capnp-futures does not need. We can instead use only the subcrates that are actually required.


dwrensha:
Looks good to me! (#199 as a previous attempt at this.)

PR #537, opened 2025-01-06
remove function pointer comparisons

dwrensha:


capnp/src/introspect.rs
```diff
@@ -272,9 +272,28 @@ pub struct RawBrandedStructSchema {
...
+        // In capnproto-c++, this method is implemented as a single pointer
+        // comparison. That's possible because C++ guarantees unique addresses
+        // for static template instantiations. A rough equivalent in Rust would be
+        // to compare the function pointers `self.field_types` and
+        // `other.field_types`, but doing so triggers a warning:
+        // > function pointer comparisons do not produce meaningful results
+        // > since their addresses are not guaranteed to be unique
+
+        if core::ptr::eq(self.generic, other.generic) {
+            return false;
```
zirconium-n, line 284: Shouldn't this return `true`?

dwrensha: Good catch! It should be negated, like this:
```rust
        if !core::ptr::eq(self.generic, other.generic) {
            return false;
        }
```
See #550.

PR #539, opened 2025-01-13
restore reason string behaviour pre 0.18

edevil:
Before 0.18.0 the reason string for an RPC error did not interpolate the kind inside. This makes sense because the type already has its own attribute.

This patch restores the original behaviour so that clients that rely on inspecting this error string can continue to operate as before.

### DIFF
--- a/capnp-rpc/src/rpc.rs
+++ b/capnp-rpc/src/rpc.rs
@@ -356,7 +356,7 @@ fn to_pipeline_ops(
 }
 
 fn from_error(error: &Error, mut builder: exception::Builder) {
-    builder.set_reason(error.to_string());
+    builder.set_reason(&error.extra);
     let typ = match error.kind {
         ::capnp::ErrorKind::Failed => exception::Type::Failed,
         ::capnp::ErrorKind::Overloaded => exception::Type::Overloaded,
--- a/capnp-rpc/test/test.rs
+++ b/capnp-rpc/test/test.rs
@@ -412,7 +412,7 @@ fn pipelining_return_null() {
         match cap.foo_request().send().promise.await {
             Err(ref e) => {
                 if e.extra
-                    .contains("Pipeline call on a request that returned no capabilities")
+                     == "remote exception: Pipeline call on a request that returned no capabilities or was already closed."
                 {
                     Ok(())
                 } else {


dwrensha:
For reference, here's the commit where the change happened: https://github.com/capnproto/capnproto-rust/commit/13ccb485e2131fc89ad9d12f1d8ee453502d3f58.
I'm trying to understand if there was a reason for it...

dwrensha:
@ariel-miculas, this pull request is proposing to revert one detail of your commit https://github.com/capnproto/capnproto-rust/commit/13ccb485e2131fc89ad9d12f1d8ee453502d3f58. Do you care to weigh in? Any reason that I shouldn't accept this change?

dwrensha:
OK, I reminded myself what's going on here.

> Before 0.18.0 the reason string for an RPC error did not interpolate the kind inside. This makes sense because the type already has its own attribute.

capnp-v0.18 added a bunch of variants to the `ErrorKind` enum (previously only the first four existed):

https://github.com/capnproto/capnproto-rust/blob/1ec4709b437ba247bc2bae02000535a03f494a33/capnp/src/lib.rs#L180-L395

However, the RPC `Exception.Type` union still only has four variants:
https://github.com/capnproto/capnproto-rust/blob/1ec4709b437ba247bc2bae02000535a03f494a33/capnp-rpc/schema/rpc.capnp#L1198-L1246

The `rpc::from_error()` function converts a `ErrorKind` into an `Exception.Type`. If the `ErrorKind` is not one of the four primary kinds, then it gets mapped to `Failed`. We do `builder.set_reason(error.to_string())` so that the extra information represented by `ErrorKind` variant is not lost; instead of just dropping that information on the floor, we allow it to influence what gets put in the `reason` string field.

Note that in the base `capnp` crate most errors do not write to the `Error::extra` field. So the change proposed in this PR would result in such errors being lost when crossing RPC boundaries.

Anyways -- I am aware that the mismatch between these error types is not ideal, and I would like to figure out a way to make it nicer.

edevil:
I think it makes sense. The case I had issues with were exceptions of type Failed, and with the linked change I have complete control over the reason string.

dwrensha:
Closing in favor of #540.

PR #546, opened 2025-02-09
Allow `ReaderSegments` to be implemented for references to unsized `ReaderSegments`

ypoluektovich:
Currently `&[&[u8]]` does not implement `ReaderSegments` despite `[&[u8]]` implementing it, because `[&[u8]]` is not `Sized`. `SegmentArray` exists to overcome this, but there's doesn't seem to be a need for forcing `Sized`, as all methods of `ReaderSegments` accept `Self` via a reference.


dwrensha:
Thanks!

PR #552, opened 2025-04-01
Fix `no_std` error trait implementation

elagil:
As far as I can tell, the `Error` trait is not fully implemented for `no_std`. This should fix it.

### DIFF
--- a/capnp/src/lib.rs
+++ b/capnp/src/lib.rs
@@ -138,10 +138,7 @@ impl core::ops::AddAssign for MessageSize {
 pub struct NotInSchema(pub u16);
 
 impl ::core::fmt::Display for NotInSchema {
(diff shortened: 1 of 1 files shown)


dwrensha:
Ah, I suppose that https://github.com/rust-lang/rust/pull/99917 made this possible.

elagil:
I guess bumping the rust version for the package would be required. Is that acceptable?

dwrensha:
I added the "Breaking Change" label to this PR. We should include this when we bump the version to v0.21.0.

dwrensha:
rebased onto master in 7f5c15dd14cd9d4f4b5093dc90e04d7c0664ecd5

elagil:
Great!

PR #553, opened 2025-04-05
Expand ReaderSegments implementations

ypoluektovich:
The motivation for this change is: I wanted to call `Reader::new` with a  `Vec<Vec<u8>>` as backing storage, without manually converting the vectors to slices.

To that end, the main change of this PR is that I implement `ReaderSegments` for vectors of `AsRef<[u8]>`.

I initially tried to make a fully generic implementation for `AsRef<[AsRef<[u8]>]>`, but Rust's lack of specializations interfered. However, one of the remaining artifacts of that attempt is the replacement of `impl ReaderSegments for [&[u8]]` with `impl<I: AsRef<[u8]>> ReaderSegments for [I]` — in human terms, for a slice of anything you can easily convert to `&[u8]`. I include that in this PR too.

I also noticed that `is_empty` was using the sub-optimal default implementation in these impls, so I correct that as well.

Finally, I add a bit of documentation to the impl for `&ReaderSegments`, to demonstrate (and enforce through doctests) that it is (still) possible to conveniently create a Reader from a static slice-of-slices.

### DIFF
--- a/capnp/src/message.rs
+++ b/capnp/src/message.rs
@@ -172,6 +172,16 @@ pub trait ReaderSegments {
     }
 }
 
+/// Allows stacking of references to a `ReaderSegments`.
+///
+/// This is especially useful with the implementation for slices, as it allows treating
+/// `&[&[u8]]` as `ReaderSegments`, e.g., to construct a `Reader`:
+/// ```
+/// # use capnp::message::ReaderSegments;
+/// use capnp::message::Reader;
+/// let slice_of_slices: &[&[u8]] = &[&*b"some data", &*b"more data"];
+/// let _ = Reader::new(slice_of_slices, Default::default());
+/// ```
 impl<S> ReaderSegments for &S
 where
     S: ReaderSegments + ?Sized,
@@ -182,6 +192,10 @@ where
 
     fn len(&self) -> usize {
         (**self).len()
+    }
+
+    fn is_empty(&self) -> bool {
+        (**self).is_empty()
     }
 }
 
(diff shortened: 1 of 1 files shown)


ypoluektovich:
I also considered adding support for collections of `IoSlice`, which is used in vectored I/O and is therefore especially likely to be encountered by users. Unfortunately, `IoSlice` is not `AsRef<[u8]>`. I'm not sure if that's a simple oversight or if there are good reasons for it being so (perhaps having something to do with how `IoSlice` is not just a newtype but a tricky unsafe-supported alias), but this means one needs to make a decision: support IoSlice specifically, or generalize it to `Borrow<[u8]>`? But then maybe a better choice would be `Deref<[u8]>`? Anyway, I didn't feel comfortable making that choice for this library.

capnp/src/message.rs
```diff
@@ -206,14 +220,38 @@ impl ReaderSegments for SegmentArray<'_> {
...
+    fn len(&self) -> usize {
+        self.len()
+    }
+
+    fn is_empty(&self) -> bool {
+        self.is_empty()
+    }
+}
+
+impl<I> ReaderSegments for Vec<I>
```
dwrensha, line 240: To make CI pass, I think you need to hide this behind `#[cfg(feature = "alloc")]` and also fully namespace `Vec` as `alloc::vec::Vec`.

ypoluektovich: Thank you very much for the hint! The fixes seem to have worked.

PR #561, opened 2025-05-16
capnp/schema: implement raw roundtrip conversion for Enum and Struct

lucab:
This augments `EnumSchema` and `StructSchema` with `Into` implementations towards their corresponding raw inner types.
It is meant to allow consumers to:
 * roundtrip between raw and non-raw types.
 * construct relevant `TypeVariant` entries from non-raw types.
 * more easily store these entities, as the raw types are `Send`.

### DIFF
--- a/capnp/src/introspect.rs
+++ b/capnp/src/introspect.rs
@@ -1,6 +1,7 @@
 //! Traits and types to support run-time type introspection, i.e. reflection.
 
 use crate::private::layout::ElementSize;
+use crate::schema::{EnumSchema, StructSchema};
 
 /// A type that supports reflection. All types that can appear in a Cap'n Proto message
 /// implement this trait.
@@ -280,6 +281,12 @@ impl core::fmt::Debug for RawBrandedStructSchema {
     }
 }
 
+impl From<StructSchema> for RawBrandedStructSchema {
+    fn from(value: StructSchema) -> Self {
+        value.raw
+    }
+}
+
 /// Type information that gets included in the generated code for every
 /// user-defined Cap'n Proto enum.
 ///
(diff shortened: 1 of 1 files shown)


dwrensha:
Should these conversions be added via the `From` trait instead? [The docs](https://doc.rust-lang.org/std/convert/trait.From.html) say:

> One should always prefer implementing From over Into because implementing From automatically provides one with an implementation of Into thanks to the blanket implementation in the standard library.

lucab:
@dwrensha thanks for the quick feedback!

Initially I considered implementing this with `From` as it is customary, but here it was feeling a bit weird for two reasons:
 * this seems more similar to an "unwrapping inner value" methods like https://doc.rust-lang.org/std/io/struct.BufReader.html#method.into_inner.
 * as the plain-type is a wrapper around the raw-type, a `RawFoo::from(WrappedFoo)` didn't sound very logical.

There was also the other option of implementing neither `From` nor `Into`, but directly going for an inherent method called `.into_inner()`.
Eventually I settled for the current code, but I'm not strongly attached to it and can happily change it to whatever you prefer.

Aside, I see there is codecov enabled on this repo; should I try to add a test covering the raw<->wrapped roundtrip?

dwrensha:
> as the plain-type is a wrapper around the raw-type, a RawFoo::from(WrappedFoo) didn't sound very logical.

You will still be able to use `.into()`, via the standard library's blanket impl.

> can happily change it to whatever you prefer.

I prefer having the `From` impls, if only to prevent Clippy from giving me this new warning;
```
warning: an implementation of `From` is preferred since it gives you `Into<_>` for free where the reverse isn't true
   --> capnp/src/schema.rs:303:1
    |
303 | impl Into<RawEnumSchema> for EnumSchema {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into
help: replace the `Into` implementation with `From<schema::EnumSchema>`
    |
303 ~ impl From<EnumSchema> for RawEnumSchema {
304 ~     fn from(val: EnumSchema) -> Self {
305 ~         val.raw
    |
```

dwrensha:
> Aside, I see there is codecov enabled on this repo; should I try to add a test covering the raw<->wrapped roundtrip?

Add a test if you want, but I'm not going to block this PR on that.

lucab:
> I prefer having the From impls, if only to prevent Clippy from giving me this new warning;

Makes sense, updated that way.

PR #564, opened 2025-05-28
Promise from Result

juchiast:


capnp/src/capability.rs
```diff
@@ -103,6 +103,14 @@ impl<T, E> Future for Promise<T, E> {
     }
 }
 
+impl<T, E> From<Result<T, E>> for Promise<T, E> {
```
dwrensha, line 106: This needs to be feature-gated via `#[cfg(feature = "alloc")]`.

PR #567, opened 2025-06-09
fix clippy issues

b00f:
This PR fixes the clippy issues (nightly version)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,9 +149,8 @@ jobs:
             sudo apt-get install -y capnproto libcapnp-dev
 
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@master
+      - uses: dtolnay/rust-toolchain@nightly
         with:
-          toolchain: nightly-2025-05-13
           components: clippy
       - uses: actions-rs-plus/clippy-check@v2
         with:
--- a/capnp/src/schema_capnp.rs
+++ b/capnp/src/schema_capnp.rs
@@ -752,14 +752,14 @@ pub mod node {
         11 => <crate::schema_capnp::node::annotation::Owned as crate::introspect::Introspect>::introspect(),
         12 => <crate::struct_list::Owned<crate::schema_capnp::node::parameter::Owned> as crate::introspect::Introspect>::introspect(),
         13 => <bool as crate::introspect::Introspect>::introspect(),
-        _ => panic!("invalid field index {}", index),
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -149,9 +149,8 @@ jobs:
             sudo apt-get install -y capnproto libcapnp-dev
 
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@master
+      - uses: dtolnay/rust-toolchain@nightly
         with:
-          toolchain: nightly-2025-05-13
```
dwrensha, line 154: In https://github.com/capnproto/capnproto-rust/pull/458#discussion_r1412855424 we decided that it was preferable for the nightly to be pinned, so that CI does not spontaneously break.

capnp/src/schema_capnp.rs
```diff
@@ -752,14 +752,14 @@ pub mod node {
         11 => <crate::schema_capnp::node::annotation::Owned as crate::introspect::Introspect>::introspect(),
         12 => <crate::struct_list::Owned<crate::schema_capnp::node::parameter::Owned> as crate::introspect::Introspect>::introspect(),
         13 => <bool as crate::introspect::Introspect>::introspect(),
-        _ => panic!("invalid field index {}", index),
+        _ => panic!("invalid field index {index}"),
```
dwrensha, line 755: Does this change (and the others in the generated `_capnp.rs` files) mean that `regenerate-capnp-schema-code.sh` and `regenerate-rpc-schema-code.sh` no longer suffice for regenerating these files?

b00f: Please check, all have linting issues with nightly.

b00f: Since these files are used to generate Rust files in the target project, we need to mark them as exceptions in our project. This PR resolves that inconvenient issue.
Look [here](https://github.com/pactus-project/tanour/blob/fe293fb90d4f8889d7483453bf6e37956220c73a/.github/workflows/workflow.yml#L50)

dwrensha:
#570 is one way to the generated code to pass Clippy while still working on Rust 2015.

dwrensha:
Thanks for the PR. I have fixed the warnings in https://github.com/capnproto/capnproto-rust/commit/6e610e809225bb2bbb44745b688a6405c4000c71, #570, and 099f350ec13efb4e22a5151b630bb4c5fa889a33.

PR #568, opened 2025-06-23
capnpc-rust: make CLI plugin aware of `default_parent_module`

lucab:
This enhances the `capnpc-rust` CLI plugin binary in order to make the default "parent module" value dynamic and controlled through an environment flag at runtime.

It matches the behavior which is already available for `build.rs` scripts through `CompilerCommand::default_parent_module()`.
This only affects the default value, which can be still overridden by `parentModule` schema annotations as usual.

When using the Rust plugin through the `capnp` compiler, it is now possible to set the default parent module like this:
```
CAPNPC_RUST_DEFAULT_PARENT_MODULE="generated" capnp \
  compile \
  -orust:./src/generated/ \
  --src-prefix=src/schema \
  ./src/schema/poc.capnp
```

### DIFF
--- a/capnpc/src/capnpc-rust.rs
+++ b/capnpc/src/capnpc-rust.rs
@@ -28,8 +28,14 @@
 pub fn main() {
     //! Generates Rust code according to a `schema_capnp::code_generator_request` read from stdin.
 
-    ::capnpc::codegen::CodeGenerationCommand::new()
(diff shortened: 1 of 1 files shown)


dwrensha:
My hope is that if we can implement #576, then we won't need this workaround.

PR #592, opened 2025-10-17
Resolve Field type in creation

bobrik:
We have this loop to resolve arbitrary user supplied inner fields based on `foo.bar.baz` paths:

```rust
            for path in paths {
                let field = *fields.get(path).unwrap();

                if let TypeVariant::Struct(_) = field.get_type().which() {
                    builder = if builder.has(field)? {
                        builder.get(field).unwrap()
                    } else if init_inactive_fields {
                        builder.init(field).unwrap()
                    } else {
                        return Err(Error {
                            kind: ErrorKind::Failed,
                            extra: format!("field {path} exists, but is uninitialized"),
                        });
                    }
                    .downcast();
                } else {
                    return Err(Error {
                        kind: ErrorKind::NotAStruct,
                        extra: path.to_string(),
                    });
                }
            }
```

There are probably better ways to do this, but that's what we have.

If you get a flamegraph of a benchmark for this, you can see a bunch of calls into `Field::get_type()`:

image

To enumerate:

* There's a direct call: `if let TypeVariant::Struct(_) = field.get_type().which() {`
* There's a call through `builder.has()`
* There's a call through `builder.get()`

It's about 10% of on-CPU time overall in a benchmark.

It's probably expected that `Field::get_type()` would be called at least once during `Field`'s lifetime, probably more, so we should front load this cost and have the type available immediately.

With this change applied the flamegraph has less CPU time spent in type resolution:

image

This lowers on-CPU time to about 1.3%.

### DIFF
--- a/capnp/src/schema.rs
+++ b/capnp/src/schema.rs
@@ -132,6 +132,7 @@ impl From<RawBrandedStructSchema> for StructSchema {
 pub struct Field {
     proto: field::Reader<'static>,
     index: u16,
+    kind: introspect::Type,
     pub(crate) parent: StructSchema,
 }
 
@@ -141,7 +142,7 @@ impl Field {
     }
 
     pub fn get_type(&self) -> introspect::Type {
-        (self.parent.raw.field_types)(self.index)
+        self.kind
     }
 
     pub fn get_index(&self) -> u16 {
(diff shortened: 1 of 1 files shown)


dwrensha:
Makes sense to me. Thanks!

Issue #577, opened 2025-07-29
`async fn` methods for RPC Server traits


dwrensha:
## Summary

We propose adding support for `async fn` syntax in impls of generated `Server` traits.

## Motivation

Suppose we have this interface defined in a capnp schema file `example.capnp`:

```capnp
interface Foo {
  bar @0 (x :UInt16) -> (y :UInt16);
}
```

Currently, capnpc-rust generates the following code in `example_capnp.rs`:

```rust
pub mod foo {
  trait Server {
    fn bar(&mut self, foo::BarParams, foo::BarResults) -> Promise<(), Error>;
  }
  // ...
}
```

Users very often need to be able to run asynchronous code from within method bodies. The `Promise` type makes that possible; `Promise<T, E>` is essentially `Box<dyn <Future<Output=Result<T,E>> + 'static>`. A promise can be constructed like this:
```rust
Promise::from_future(async move {
  // ... do some asynchronous stuff here
})
```

However, there are some problems with this setup:

1. Rust users accustomed to `async/await` find it annoying to need to manually construct a `Promise` like this. See #168 for some discussion.

2. The `?` operator does not work inside an RPC method body without the `try_trait_v2` unstable feature. (We currently work around this with a [`pry!()`](https://github.com/capnproto/capnproto-rust/blob/58977b7ae792c413df21a75f982a406cc56a31eb/capnp-rpc/src/lib.rs#L84-L98) macro.

3. There's no way to get access to the borrowed `self` parameter from within the `async` context, because the future needs to be `'static`.

Fortunately, thanks to some new-ish [features in the Rust compiler](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/), we should now be able to address these problems.

## Solution

We will adjust the generated trait to be:

```rust
pub mod foo {
  trait Server {
    fn bar(self: Rc<Self>,
           foo::BarParams,
           foo::BarResults) -> impl Future<Output=Result<(), Error>> + 'static;
  }
  // ...
}

```

Then, user-provided implementations can use the `async fn` syntax like this:

```rust

impl foo::Server for FooImpl {
  async fn bar(self: Rc<Self>, foo::BarParams, foo::BarResults) -> Result<(), Error> {
    // ... do some asynchronous stuff here, possibly using the `?` operator
  }
}
```

[...]

PR #593, opened 2025-10-23
add support async fn in Server traits

luisholanda:
This PR changes how we generate the `Server` traits for interfaces. We now generate `async fn` compatible traits instead of methods returning `Promise`.

The methods now receive `&self` instead of `&mut self`, which causes server implementations to need `RefCell`/`Cell` for mutable state, similar to what we have for `Send` servers in other RPC systems. This is slightly different from what was discussed in #577, because we don't actually have a `Rc<impl Server>` inside `Client`, but instead `Rc<ServerDispatch>`, making the receiver a `Rc<Self>` (which would allow easier background processing spawning) would require an extra allocation, pointer chasing, and ref-counting operations.

From the changes in the examples, we can see that the code is significantly easier to understand, with `pry!` rarely being needed. The only case I truly needed to use it was when the code needed to manually create a `Promise` due to a `RefMut` preventing the use of `async fn` (which could cause double borrows). For me, this is a significant improvement in UX.

In matters of breaking changes, this, of course, breaks every pre-existing implementation, but it should be an easy migration. In the crates themselves, `capnp::capability::Server` needed to be changed to receive `Rc<Self>` instead of `&mut self`, not only the `&mut` isn't necessary anymore but also receiving a `Rc<Self>` (which we already have where it is called) also allow us to make the generated methods for `Server` traits non-'static, which makes the `async fn` syntax significantly more useful. Additionally, certain methods in `CapabilityServerSet`, that exposed the internal `Rc<RefCell<Dispatcher>>` from `Client`, needed to be changed to return `Rc<Dispatcher>`. These changes in the crates were the main reason why there isn't a switch for old/new code generation methods, as they are incompatible with each other.

Also, given that we're using `async/await` in the generated code, they can't be compiled in editions 2015/2018 anymore. It may be possible to support them with some added complexity in the generated code, but I don't think it is worth it.

Closes: #577


luisholanda:
It seems I forgot to check some stuff locally! My bad :(

I'll fix CI tomorrow (currently kinda of late here 😅).

dwrensha:
> Strictly from an application writers perspective, I'd rather avoid having to deal with Rc, RefCell etc, if I can.

What changes in this PR is that now your `Server` impls need to bring their own interior mutability. So you may need to add some `Cell` or `RefCell`  wrappers. I don't think you'll need to add any `Rc`, unless I misunderstand something.

dwrensha:
> My motivation is that I'm using an io_uring-based runtime for my project and need a custom VatNetwork with the lowest possible overhead.

I think the improvements in this PR are important enough that we should not block them on further speculative changes.

luisholanda:
@dwrensha, I've updated the PR, fixed the CI, and applied your suggestions to the code.

PR #595, opened 2025-10-28
&self -> Rc<Self> in RPC methods

dwrensha:
Follows the suggestions in https://github.com/capnproto/capnproto-rust/issues/577#issuecomment-3454365512

### DIFF
--- a/capnp-futures/Cargo.toml
+++ b/capnp-futures/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "capnp-futures"
-version = "0.22.0"
+version = "0.23.0-alpha"
 authors = [ "David Renshaw <drenshaw@gmail.com>" ]
 license = "MIT"
 
@@ -12,7 +12,7 @@ edition = "2021"
 keywords = ["async"]
 
 [dependencies]
-capnp = { version = "0.22.0", path = "../capnp" }
+capnp = { version = "0.23.0-alpha", path = "../capnp" }
 
 [dependencies.futures-channel]
 version = "0.3.0"
@@ -30,7 +30,7 @@ default-features = false
 features = ["executor"]
 
 [dev-dependencies]
-capnp = { version = "0.22.0", path = "../capnp", features = ["quickcheck"] }
+capnp = { version = "0.23.0-alpha", path = "../capnp", features = ["quickcheck"] }
 quickcheck = "1"
 
 [lints]
--- a/capnp-rpc/Cargo.toml
+++ b/capnp-rpc/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "capnp-rpc"
-version = "0.22.0"
+version = "0.23.0-alpha"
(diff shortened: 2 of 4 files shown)


dwrensha:
@nivereno @abronan @andrewbanchich @kjuulh @rschulman @luisholanda I'm interested in your thoughts on this.

capnpc/src/codegen.rs
```diff
@@ -2592,7 +2592,7 @@ fn generate_node(
                 if result_id != STREAM_RESULT_ID {
                     dispatch_arms.push(
                         Line(fmt!(ctx,
-                                  "{ordinal} => {capnp}::capability::DispatchCallResult::new({capnp}::capability::Promise::from_future(async move {{ (**this).{}({capnp}::private::capability::internal_get_typed_params(params), {capnp}::private::capability::internal_get_typed_results(results)).await }}), false),",
+                                  "{ordinal} => {capnp}::capability::DispatchCallResult::new({capnp}::capability::Promise::from_future(this.{}({capnp}::private::capability::internal_get_typed_params(params), {capnp}::private::capability::internal_get_typed_results(results))), false),",
```
nivereno: Pretty sure this can be replaced with just self

capnpc/src/codegen.rs
```diff
@@ -2647,13 +2647,13 @@ fn generate_node(
                     // It's a streaming method.
                     dispatch_arms.push(
                         Line(fmt!(ctx,
-                                  "{ordinal} => {capnp}::capability::DispatchCallResult::new({capnp}::capability::Promise::from_future(async move {{ (**this).{}({capnp}::private::capability::internal_get_typed_params(params)).await }}), true),",
+                                  "{ordinal} => {capnp}::capability::DispatchCallResult::new({capnp}::capability::Promise::from_future(this.{}({capnp}::private::capability::internal_get_typed_params(params))), true),",
```
nivereno: Our implementation was devised pre DispatchCallResult existing(We have this function as async with allow async in traits and simply await the result), but something I don't like here is that both the local::Client::call and this has a Promise::from_future, which boxes and pins the future, seems like unnecessery overhead for each call

dwrensha:
@nivereno thanks for the suggestions. I think making `extends` more ergonomic is an interesting discussion, but outside of the scope of this PR. I want to quickly make a release with the new `Rc<Self>` stuff, to minimize the pain of upgrading. (If folks haven't upgraded to 0.22 yet we can advise them to skip it.)

capnp-rpc/src/lib.rs
```diff
@@ -379,7 +379,7 @@ where
     C: capnp::capability::FromServer<S>,
 {
     capnp::capability::FromClientHook::new(Box::new(local::Client::new(
-        <C as capnp::capability::FromServer<S>>::from_server(s),
+        <C as capnp::capability::FromServer<S>>::from_server(Rc::new(s)),
```
nivereno: You should add a new_client_from_rc convenience function here and to the CapabilityServerSet that takes Rc\<S\>

Issue #68, opened 2016-11-22
Consider merging with Serde


erasmospunk:
The [Sedre](https://serde.rs/) library supports many serialization formats but not capnproto.

One useful feature is the [transcode](https://serde.rs/transcode.html) functionality where it is possible to convert from one format to the other.

dwrensha:
Hi!

My impression is that Cap'n Proto would not be a very good fit for Serde integration, mainly because Cap'n Proto wants to define data types in schemas external to Rust code. I have [an old post](https://dwrensha.github.io/capnproto-rust/2014/06/01/libserialize-traits.html) on this topic. (The `libserialize` discussed there is now called [`rustc_serialize`](https://github.com/rust-lang-nursery/rustc-serialize), if I remember correctly.)

Do you have a specific use case in mind?

burdges:
I'd think cap'n proto could generate the serde attributes from its schema, maybe useful if someone must work around company policies like always use json or xml for such and such.

erasmospunk:
Good point about the schema, I initially thought that capnpc-rust could generate the needed `struct`s with Sedre's annotations & any runtime code. Not sure if this is simple, especially considering that Sedre does not support protocol buffers.

My use case is converting from JSON to Capnproto as an internal representation for microservices. Using the Sedre lib could potentially make the transcoding with less code.

If you think that this is something that does not make sense for this project, feel free to close this issue.

dwrensha:
I do think think that it would be good to have JSON <--> Cap'n Proto support in capnproto-rust, but I'm not sure that it would make sense to involve Serde. I would probably start by looking at how [capnproto-c++ supports JSON encoding and decoding](https://github.com/sandstorm-io/capnproto/tree/b6d7079ac0cba1cf7003bbc6a889374d45a10087/c%2B%2B/src/capnp/compat).

erasmospunk:
Thanks for feedback, closing this issue...

dwrensha:
Today I came across [serde-protobuf](https://github.com/dflemstr/rq/tree/master/serde-protobuf). I'm not sure exactly what it does, but it seems relevant to this discussion.

piaoger:
Do you have a plan to support JSON encoding and decoding in capnproto-rust?

dwrensha:
@piaoger It would be good to have, but I don't have any immediate plans to work on it.

[...]

PR #600, opened 2025-12-03
RFC/WIP: Introduce capnp-json for JSON code compatible with c++ default capnproto implementation

puremourning:
As discussed in #68, I'm sharing this very rough cut of JSON codec support for input.

Some assumptions I made, which influenced the implementation, but could be changed depending on your view:

- avoid third party crates: json parser/encoder rolled from scratch, base64/hex encoders rolled from scratch
- should be its own crate in the workspace (per comment in #68)
- std is ok for this crate
- perfect is the enemy of good (just kidding, or rather trying to soften the blow, as this implementation is not art)

What's supported:

* Compatible with the _default_ encoding provided by the upstream c++ implementation I believe
* Encode and Decode to/from String
* Annotations provided by upstream: `flatten`, `discriminator`, `name`
* `base64`, `hex` and `array of numbers` encodings for `Data` fields.
* a smattering of tests. coverage is probably not yet sufficient.

The best doc on the actual encoding is the json.capnp which is taken directly from upstream. We don't use all of it, but I kept it unmodified anyway.

What's _not_ supported, and what deviates from the upstream implementation:

* Any way to customise the encoding/decoding. The c++ implementation has a sort of codec plugin which allows users to supply arbitrary codecs for fields.. This is probably most useful for Data fields.
* Speed. This is not an optimised parser, serialiser or anything. It's the simplest way to just do something that works.
* Arbitrary "JsonValue" types in messages. This is due to a difference in the way the codecs work. The c++ implementation uses a JsonValue (capnp) as an intermediary format for any value type whether encoding or decoding. This relies on Orphans (which we don't support) and seems to me to be an unnecessary overhead. In this implementation we go direct from Reader -> JSON string. For decoding, we still need an intermediary format, so we use a JsonValue of our own (a Rust enum) and thus JSON string -> JsonValue enum -> Builder.
* There's a lack of comments and general polish; to be done if we agree on the bigger picture items. 
* Security safety rail; there is no protection over blowing out the stack, due to a maliciously crafted JSON document


capnp-json/src/decode.rs
```diff
@@ -0,0 +1,804 @@
...
+                    field_meta.name
+                )));
+            };
+            Ok((*field_value.as_str()).into())
+        }
+        capnp::introspect::TypeVariant::Enum(enum_schema) => match field_value {
+            JsonValue::String(field_value) => {
+                let enum_schema = capnp::schema::EnumSchema::new(enum_schema);
+                let Some(enum_value) = enum_schema.get_enumerants()?.iter().find(|e| {
+                    // FIXME: this is naive, enum values can be renamed using
```
puremourning: Remove comment, it's handled

Issue #602, opened 2025-12-06
Multiple Functions of the same name


lenianiva:
```capnproto
interface Inner {
	stop @0 () -> ();
	# Shutdown the server
}
interface Outer extends(Inner) {
	stop @0 () -> ();
	# Shutdown the server
}
```
If there are multiple functions of the same name but in different interfaces, the protocol will pass transpilation, but the generate rust file will fail to compile:
```
error[E0034]: multiple applicable items in scope
    --> .../protocol_capnp.rs:607:66
     |
 607 |         1 => ::capnp::capability::DispatchCallResult::new(server.stop(::capnp::private::capability::internal_get_typed_params(params), ::capnp::private::capability::internal_get_typed_results(results)), false),
     |                                                                  ^^^^ multiple `stop` found
     |

```

PR #608, opened 2025-12-24
fix: disambiguating function calls for interface extending

Oyami-Srk:
Fixes https://github.com/capnproto/capnproto-rust/issues/602

Also introduces regression test for this fix.

Thanks @lenianiva for discovering this issue and @dwrensha for providing the solution.


dwrensha:
Thanks!

PR #638, opened 2026-03-29
Use NonNull in Allocator

stepancheg:
This matches API of std allocator https://doc.rust-lang.org/std/alloc/trait.Allocator.html

But more broadly speaking, in capnp-rust codebase, some pointers are used with zero is information, and in other places pointer being zero is not valid. might be better to convert all the pointers to `NonNull`, and use `Option` where null value is expected.

All pointers start with allocator, so this PR.


dwrensha:
Rebased and landed as https://github.com/capnproto/capnproto-rust/commit/9759b10fa212dcddec54b4e4670d067cbd776e39

PR #651, opened 2026-04-12
.offset() -> .add()

stepancheg:
Use `.add()` instead of `.offset()` where argument is unsigned so
- more clear what is happening there
- fewer casts in code

### DIFF
--- a/capnp/src/private/layout.rs
+++ b/capnp/src/private/layout.rs
@@ -446,7 +446,7 @@ mod wire_helpers {
                 let amount_plus_ref = amount + u32::try_from(POINTER_SIZE_IN_WORDS).unwrap();
                 let (segment_id, word_idx) = arena.allocate_anywhere(amount_plus_ref);
                 let (seg_start, _seg_len) = arena.get_segment_mut(segment_id);
-                let ptr = seg_start.offset(word_idx as isize * BYTES_PER_WORD as isize);
+                let ptr = seg_start.add(word_idx as usize * BYTES_PER_WORD);
 
                 //# Set up the original pointer to be a far pointer to
                 //# the new segment.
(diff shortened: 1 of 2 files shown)


capnp/src/private/layout.rs
```diff
@@ -160,7 +160,7 @@ impl WirePointer {
     #[inline]
     pub unsafe fn target(ptr: *const Self) -> *const u8 {
         let this_addr: *const u8 = ptr as *const _;
-        unsafe { this_addr.offset(8 * (1 + (((*ptr).offset_and_kind.get() as i32) >> 2)) as isize) }
+        unsafe { this_addr.add(8 * (1 + (((*ptr).offset_and_kind.get()) >> 2)) as usize) }
```
dwrensha: The shift-right operator does different things on `i32` and `u32`. The behavior on `i32` is the correct behavior here, because if `offset` is negative, then we don't want the shift to make it not negative.

Compare to [the C++ implementation](https://github.com/capnproto/capnproto/blob/4a47e15129440773a6b765b84394bc699c7531e8/c%2B%2B/src/capnp/layout.c%2B%2B#L144), or just to `target_from_segment()`, the next function in this file.

stepancheg:
I reverted this line, but it seems like some test is missing here.

stepancheg:
This PR is needed to enable `deny(clippy::cast_possible_wrap)` which is another dangerous thing this crate does.

dwrensha:
> I reverted this line, but it seems like some test is missing here.

It's tricky to get coverage of this particular code path, because the `target()` function only gets invoked on "trusted" messages that are part of generated code. Such messages do not have backwards pointers.

Instead, I factored out the pointer arithmetic into a new method, so that it gets covered by the existing tests: https://github.com/capnproto/capnproto-rust/commit/4aa1fccf11e95e58c1c037cdfd4f0159b76ef20a

PR #657, opened 2026-04-29
Reproduce and fix bug related with `-> stream` not being queued

gabivlj:
Awaiting a -> stream call only means flow-control has room to send more data. It does not guarantee the receiver has finished processing that call. A following non-streaming call, like end() in the bytestream example, could run before earlier streaming write() calls completed.

In our real usage this sometimes surfaced as:
`Error { kind: MessageContainsOutOfBoundsPointer, extra: "" }`
In this repro, I was not able to reproduce that exact out-of-bounds error. The repro reliably shows the underlying ordering issue instead: end() completes before all bytes have arrived.

The changes have been driven by GPT 5.5, the main fixes are not depending on the runtime processing tasks in a FIFO way, and having our own queue of call dispatches in `local.rs`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ members = [
     "capnp-rpc/examples/calculator",
     "capnp-rpc/examples/pubsub",
     "capnp-rpc/examples/streaming",
+    "capnp-rpc/examples/bytestream",
     "capnp-rpc/examples/reconnect",
     "capnp-rpc/test",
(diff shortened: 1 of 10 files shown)


dwrensha:
Thanks! Closing in favor of #660.

gabivlj:
Wow, thanks so much for the fast review, cleaning up + merging @dwrensha! Much appreciated.

PR #663, opened 2026-05-21
Capability pipelining with a motivating example

sjep:
Say we write a generic wrapper interface that is intended to expose an inner capability:

```
interface SharedSecretAuthenticated(T) {
# This interface is generic over an underlying protected interface T
  authenticate @0 (sharedSecret :Text) -> (authenticated :T);
}
```

The server works great, the inner type must be `Owned`, but the inner capability can be wrapped by the server impl via the `Owned` client type.

The client also works well when using `send().promise`. First the client unwraps the internal Owned client, then it can use the internal capability.

The problem comes with pipelining the client. It would be great if the wrapper interface could be pipelined with use of the inner capability so that only one `send()` need be awaited. This just requires that `Client` implement `FromTypelessPipeline`.

I've included a motivating example that doesn't compile without the impls for `FromTypelessPipeline`.

### DIFF
--- a/capnp/src/capability.rs
+++ b/capnp/src/capability.rs
@@ -312,6 +312,7 @@ pub trait FromTypelessPipeline {
     fn new(typeless: any_pointer::Pipeline) -> Self;
 }
 
+#[cfg(feature = "alloc")]
 impl<T: FromClientHook> FromTypelessPipeline for T {
     fn new(typeless: any_pointer::Pipeline) -> Self {
         Self::new(typeless.as_cap())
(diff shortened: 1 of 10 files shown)


capnp/src/capability.rs
```diff
@@ -312,6 +312,12 @@ pub trait FromTypelessPipeline {
     fn new(typeless: any_pointer::Pipeline) -> Self;
 }
 
+impl<T: FromClientHook> FromTypelessPipeline for T {
```
dwrensha, line 316:
```suggestion
#[cfg(feature = "alloc")]
impl<T: FromClientHook> FromTypelessPipeline for T {
```

dwrensha:
squashed and merged as https://github.com/capnproto/capnproto-rust/commit/a9c60a2ec52a382bba6b81b698467d8ed0f568fc

Issue #582, opened 2025-08-12
Generated program elements should include Rustdoc from Cap'n Proto schema


toddATavail:
I'm open to many possibilities for how this could be realized, but the generator ought to take any element docs from Cap'n Proto schema files and emit them appropriately in the generated code. I realize that a fully "literate" approach that massages wording after the fact would be quite challenging without appealing to an unacceptably heavy machine learning model, but hopefully pasting the source docs into the output verbatim wouldn't be too difficult. Even that much would add a lot of value.

Let's use the following Cap'n Proto schema snippet for reference:

```capnp
struct Point {
   # A 2D point.

    x @0 :Float64;
    # The x-coordinate.

    y @1 :Float64;
    # The y-coordinate.
}
````

The `struct` comment (`"A 2D point."`) could be added as Rustdoc to the generated `Builder` and `Reader`. The field comments (`"The x-coordinate."`, `"The y-coordinate."`) could be added to the getters and setters of `Builder`s and `Reader`s.

The obvious use cases are `cargo doc` and development tooling (e.g., IDEs, LSP servers, …). A substantial amount of effort might go into writing good comments in the Cap'n Proto schema files, and it's a shame to lose that value after code generation. Currently, a developer needs to keep open the original schema files while developing against the Rust bindings in order to see the original documentation, and it would be nice to have the docs available from the tooling while coding.

Issue #165, opened 2020-03-05
Using comments as docstrings


zdb999:
I've been playing around with the Sandstorm bindings, and often I've had problems interpreting the methods from the generated docs. I end up bouncing back and forth between the capnp spec and the generated Rust code several times when trying to understand each a relatively simple operation. It would be really helpful if the generated Rust code incorporated the documentation in the .capnp files. Would it be possible for a macro to convert the comments in .capnp files into Rust docstrings? Would it make sense to do so?

dwrensha:
Yes, this would be possible: https://github.com/capnproto/capnproto/blob/a68afd9e78839a7cf2a83b5e49134c5161f7a2d2/c%2B%2B/src/capnp/schema.capnp#L498-L500

And yes, I think it would be a good idea.

zdb999:
I've been playing around with this, and it appears that the sourceInfo list is always empty. (I'm calling `gen.request.get_source_info()?`) Do you know why this might be occurring?

dwrensha:
What version of the `capnp` tool do you have installed? (What's the output of `capnp --version`?)

zdb999:
It's version 0.6.1.

dwrensha:
Ah, I think the `source_info` stuff was added in 0.7.

schreter:
I also vote for supporting this. We have capnp files with extensive documentation, which isn't propagated to the generated files in Rust. Our C++ version of the code uses a special generator transforming capnp IDL files into Doxygen comments, so we have nice combined reference docs for C++ code. It would be nice to have this for pure Rust development as well.

dwrensha:
If the schema file has comments on struct fields, where in the generate code should those comments live? Each field maps to many methods split among the `Reader` and `Builder` objects, and duplicating the comments seems wasteful. Maybe the canonical place would be for them to go on the `Reader::get_*()` methods?

schreter:
Right... The C++ version we have is actually generating dummy headers with C++ structs (i.e., without accessors, plain fields, and pointers for referenced structs/lists). This is used in our Doxygen documentation generator to generate the docs and only for this.

In Rust, generated code contains top-level objects like modules for each capnp `struct` or enums for each capnp `enum`. That's indeed easy to document. Also enum members are easy to document. Struct fields are the problem.

There are several potential options how to handle `struct` fields:
[...]

PR #665, opened 2026-06-08
feat: add docstrings to generated rust code

ksil:
First-time contributor to the repo here! Please feel free to be picky and let me know if there are any cultural or stylistic things to change!

This is a (hopefully) small change to read the source info from the schema and use it to generate rust doc strings, which should address https://github.com/capnproto/capnproto-rust/issues/165 and https://github.com/capnproto/capnproto-rust/issues/582.

Struct-level comments are added to the module ~~as well as Owned, Reader, Builder, and Pipeline structs. This is potentially overkill, but I thought it was kind of nice for tooltips in IDEs and at various levels of the generated cargo docs. Happy to debate it, though.~~

Field-level comments are added to getters and setters in Builders and Readers. File-level comments are added at the top of the generated rust file (and are not included in cargo docs because `//!`-type comments can easily break builds). Enum value comments are added before every value. Interface comments are added for both Client and Server methods. Constants feature comments before their definitions. Annotation comments are added to the module blocks.

Also, I added .idea to the list of gitignored things for those that use RustRover like me.


ksil:
@dwrensha assuming you're the right person to review this, but let me know if I should loop in others! I don't think I have permission to add reviewers or launch CI workflows

dwrensha:
Thanks for the PR. My main hesitation here is that this seems like a lot of duplication and could bloat the size of the generated files. Sometimes documentation for individual fields can span multiple paragraphs. (See e.g. [this sandstorm schema](https://github.com/sandstorm-io/sandstorm/blob/a97cf3ee19d3bf2761cd597583ca5d998de425d8/src/sandstorm/grain.capnp).) If each such comment gets duplicated three times, it seems like things could get out of hand.

ksil:
> Thanks for the PR. My main hesitation here is that this seems like a lot of duplication and could bloat the size of the generated files. Sometimes documentation for individual fields can span multiple paragraphs. (See e.g. [this sandstorm schema](https://github.com/sandstorm-io/sandstorm/blob/a97cf3ee19d3bf2761cd597583ca5d998de425d8/src/sandstorm/grain.capnp).) If each such comment gets duplicated three times, it seems like things could get out of hand.

I just removed the structure comments on Owned, Reader, Builder, and Pipeline, so now they're just attached to the associated Rust module. Not sure what to do about fields, though, because they are nice to see in the IDE on both the getters and setters and aren't tied to a module. What do you think?

ksil:
Hi @dwrensha, let me know if we can discuss anything or if there's anything else I can change! Appreciate the help.

ksil:
Just following up here @dwrensha if there's anything I can do to move this PR along! Apologies for bothering you about it; just wanted to try to contribute a useful change to Rust capnproto here that several of us at my company would love to see

dwrensha:
The 4x duplication for field doc comments is what makes me hesitate here. I wonder whether it would make sense to hoist the field comments into the module-level struct comment, so that they appear exactly once. And perhaps getters and setters could have comments that somehow link to the module-level comment.

ksil:
It's pretty common practice across other IDL code generators (like Protobuf in Java and Thrift) to intentionally duplicate field comments across every accessor to prioritize developer experience in IDEs. This duplication only lives in the autogenerated Rust files, which people generally should not check into repos and should not reference directly. These are also comments, which negligibly affect compiler speed compared to other AST elements. Bloat in non-autogenerated files would be a different story. I opt for prioritizing IDE hover support over DRYness in autogenerated build files. Indirection to other module-level comments will most likely make the IDE experience more challenging than necessary.

PR #671, opened 2026-07-18
Expose a public broken-capability constructor

lthibault:
Building a capability membrane (method-level attenuation à la C++ `membrane.h`) on top of `ClientHook` requires a "broken" client/request/pipeline for the deny path — a cap that returns a given error from every operation. `capnp-rpc` already has exactly this in its `broken` module, but it's `pub(crate)`, so every downstream membrane re-copies ~120 lines of it.

This PR exposes a minimal public constructor and keeps the internals private. C++ capnp exposes `newBrokenCap` publicly for the same reason; membrane-style interposition is impractical downstream without it.

Includes a unit test that the returned client errors on `call`/`new_call` and pipelined access.

Opening as **draft** to gauge interest before polishing.

### DIFF
--- a/capnp-rpc/src/lib.rs
+++ b/capnp-rpc/src/lib.rs
@@ -392,6 +392,21 @@ where
     )))
 }
 
+/// Creates a "broken" capability that returns the given error from every operation.
+///
+/// The returned client fails with a clone of `error` on every method call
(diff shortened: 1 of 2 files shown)


lthibault:
cc @dwrensha

dwrensha:
Seems fine to me.

lthibault:
Great!  Ready for review.

Issue #536, opened 2025-01-06
 Compile warning on rustc 1.86.0-nightly: function pointer comparisons do not produce meaningful results


zirconium-n:
```
warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
   --> capnp/src/introspect.rs:275:55
    |
275 |         core::ptr::eq(self.generic, other.generic) && self.field_types == other.field_types
    |                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the address of the same function can vary between different codegen units
    = note: furthermore, different functions could have the same address after being merged together
    = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>
    = note: `#[warn(unpredictable_function_pointer_comparisons)]` on by default

```
```
> rustc --version
rustc 1.86.0-nightly (b3b368a18 2025-01-05)
> git rev-parse HEAD
1ec4709b437ba247bc2bae02000535a03f494a33
```

dwrensha:
Thanks for the report. I think this means that we need to remove the `PartialEq` and `Eq` impls in `introspect.rs`. I started a draft PR to do that: #537.

dwrensha:
Unfortunately, I needed to revert to the old impl in #550.  We're now suppressing the warning, and nothing internal to capnp depends on the possible unpredictable behavior. But still downstream users might be unpleasantly surprised if they use equality on types, so I opened #551 to track further progress on this issue.

PR #675, opened 2026-08-01
Support Eq and Hash for Field and StructSchema

puremourning:
In order to provide support for overriding JSON field codecs in parity with the upstream C++ Cap'n Proto reference implementation, we need a way to store and compare Field instances. The C++ implementation uses the pointer address of the StructSchema and the index as a stable Field identity, so we do the same.

---

This is for the implantation of https://github.com/puremourning/capnp-json/pull/2 and of course cannot be done there due to the orphan rule.

No AI was used.


capnp/src/schema.rs
```diff
@@ -124,6 +124,26 @@ impl From<RawBrandedStructSchema> for StructSchema {
     }
 }
 
+impl std::cmp::PartialEq for StructSchema {
+    fn eq(&self, other: &Self) -> bool {
+        // The identity of the schema is the pointer value of the 'static actual
+        // struct deifnition. There can only be one such definition. This
+        // matches the C++ implementation.
+        std::ptr::eq(
```
puremourning: just thought. I need to check the semantics with generic types.
If I read the c++ code correctly, then the overrides are branding-invariant, but I think I need to double-check.

puremourning: actually, this _is_ wrong. c++ hashes the RawBrandedStructSchema not the RawStructSchema.
need to rethink.

puremourning: Done - we now compare the full RawBrandedStructSchema.

puremourning: Despite my thoughts/comments and the tests working. A `miri` test does not - because miri seems to intentionally make these fn pointer comparisons fail.

dwrensha: This looks like the issue we ran into in https://github.com/capnproto/capnproto-rust/issues/536. Looks like `fn_addr_eq` suppresses the warning, but does not avoid the potentially unpredictable behavior.

puremourning: I did some research and yeah the behaviour is _sound_ and _almost certainly correct for almost certainly all usages_ (except maybe across a DLL boundary), but it is relying on _unspecified_ behaviour.

This is irksome. We _could_ add a `::core::any::TypeId` into `RawBrandedStructSchema`, instantiated for the `Owned` or `Owned<T,...>` and require that `Introspect : 'static` . This would obviously be a breaking change, though almost certainly the only things that create `RawBrandedStructSchema` and `impl Introspect` are the `capnpc-rust` compiler which we can control anyway.

I'll try that out and see if it works...

dwrensha: I'm willing to make a breaking change if that's what it takes to do this the right way.

puremourning: OK, nice. I've pushed the change to use a TypeId. I'm going to do a bit more careful reading of [the docs](https://doc.rust-lang.org/beta/std/any/struct.TypeId.html) but I can confirm this change passes the tests, and my capnp-json tests including under `miri`. I'd appreciate it if we can trigger the CI for this project to confirm the various. profiles there too.

dwrensha: Did you end up not needing to so `Introspect : 'static`?

dwrensha: After this change, can we get rid of `Type::loose_equals()`, in favor of a precise equality?

puremourning: Indeed - it's actually only the type parameters to generated Owned<P,Q,R,S,T> that need to be 'static for `TypeId::of::<Owned<P,Q,R,S,T>>()` to be valid. So all I had to do was add the `'static` bound to those in more places - we already had them in some places, and it's always true for the generated types.

puremourning:
CI failure due to `std::hash` rather than `core::hash` fixed.

capnp/src/schema.rs
puremourning: > After this change, can we get rid of Type::loose_equals(), in favor of a precise equality?

Hmm, good question. Let me look into it - have to confess, I actually wasn't aware of loose_equals.

puremourning: Turns out, yes we can - indeed we can reinstate the auto-derive of PartialEq and Eq on Type and BaseType and even auto-derive Hash. I'll add it in another commit and add some tests.

puremourning: done

